Issue #118, opened 2026-04-08
"Show context" in filtered Trace view


sfrankle:
## Story
As a user, I want to peek at surrounding entries while viewing a filtered Trace so that I can answer "what else was going on at this moment?" without losing my focused view.

## Acceptance criteria
- [ ] When a Focus filter is active, individual entries show a "Show context" button
- [ ] Tapping it temporarily reveals surrounding entries (±2 hours) without clearing the filter or losing scroll position
- [ ] Contextual entries appear in a visually muted style to distinguish them from filtered results
- [ ] Dismissing context returns to the normal filtered view

## Notes
This makes the Focus filter genuinely useful for health conversations — you can see your knee entries and immediately check what else was logged around that time.

PR #154, opened 2026-04-16
"Show context" in filtered Trace

### LINKED ISSUE #137: "Show context" in filtered Trace
## Summary
When a Focus filter is active in Trace, add a "Show context" button to individual entries that temporarily reveals surrounding entries (±2 hours) in a muted style.

## Acceptance criteria
- [ ] "Show context" button appears on entry rows only when a Focus filter is active
- [ ] Tapping reveals entries within ±2 hours of the selected entry, without clearing the filter or losing scroll position
- [ ] Contextual entries rendered in a visually muted style to distinguish from filtered results
- [ ] Dismissing context returns to normal filtered view
- [ ] Tests: context entries fetched for correct time window; muted style applied; dismiss restores filtered view

## Notes
Files: `app/(tabs)/trace.tsx`, `src/lib/db/queries.ts` (new `getContextEntries(db, entryId, windowMinutes)` query)

Blocked by: #136

sfrankle:
## Summary

- When a Focus filter is active in Trace, entry rows now show a "Show context" button that reveals surrounding entries (±2 hours) in a muted style without clearing the filter or losing scroll position
- Implements `getContextEntries` query in the data layer and `ContextEntryRow` component in the Trace screen; context is cleared automatically when the active filter changes
- Covers the new behaviour with unit tests (query time-window correctness, muted style, dismiss) and a Maestro E2E flow

Closes #137
Contributes to #118

## What to review

- `src/lib/db/queries.ts` — `getContextEntries` and the extracted `collapseTraceRows` helper; verify the ±2-hour window boundary logic
- `app/(tabs)/trace.tsx` — `ContextEntryRow`, `contextMap` state, `handleShowContext`/`handleHideContext`, and the `useEffect` that clears context on filter change; confirm the muted opacity approach fits the design system
- `maestro/flows/trace/focus-filter-show-context.yaml` — E2E flow covering the full show/hide interaction

## Testing

- New test file `src/lib/db/__tests__/getContextEntries.test.ts` verifies correct time-window scoping and that the focal entry is excluded
- `src/__tests__/screens/TraceScreen.test.tsx` extended: "Show context" button only appears with an active Focus filter, muted style applied to context rows, dismiss restores filtered view
- Maestro flow `trace/focus-filter-show-context.yaml` covers the full interaction end-to-end

## Checklist
- [x] Acceptance criteria met
- [x] CI passing


sfrankle (commented):
## Simplify Pass

**Fixed:**
- Replaced `useEffect` clearing `contextMap` on filter change with inline `setContextMap(new Map())` in `handleFocusPillPress` — eliminates a two-render stale flash and is clearer about causation
- Added `setContextMap(new Map())` inside `useFocusEffect` so context is intentionally cleared on navigation-away (previously only filter change flushed it)
- Added `contextMap.has(entry.id)` guard at the top of `handleShowContext` to explicitly prevent concurrent re-fetch if tapped twice before the first request resolves
- Removed now-unused `useEffect` import

**Skipped (out of scope or false positives):**
- `EntryRowBase` unified component: merging `ContextEntryRow` and `EntryRow` into a single muted-prop component is a real maintainability win but a non-trivial design refactor — out of scope
- Parameter sprawl on `EntryRow`: valid concern but requires a design decision (contextSlot prop vs sub-component decomposition)
- `SOURCE_TYPE` constant: broader codebase refactor
- `contextMap` LRU cap: single-user local app, low severity today
- `context-entries-{id}` wrapper View: actively used in 5 test assertions — keeping it

sfrankle (commented):
## Product Vision Review

**Verdict:** CONCERNS

### User Story Fulfillment

The implementation satisfies every acceptance criterion from issue #118 and #137: the button is gated behind an active Focus filter, context entries appear inline in a muted style, scroll position is preserved, and dismissing restores the filtered view. The ±2-hour window and the useEffect that clears context on filter change are both solid. Functionally, this is correct.

### UX Principles

No scores, streaks, or pressure language anywhere in the diff. The feature surfaces surrounding data without prescribing anything. Local-only — no network calls introduced. These are all fine.

The one structural concern: when `getContextEntries` returns an empty array (no surrounding entries within ±2 hours), the button is tapped but nothing appears — no feedback at all. The code renders `contextEntries.length > 0` to show rows, and the Hide button only renders when `contextEntries !== undefined && contextEntries.length > 0`. So the user taps "Show context", nothing happens visually, and no Hide button appears. This is a dead end. The user has no way to know whether the query ran, found nothing, or silently errored. It does not need to feel like failure — but it does need to feel like something: a quiet inline note such as "Nothing else logged around this time" would close the loop without pressure.

### Tone and Copy

"Show context" and "Hide context" are clear, functional-register copy — appropriate for a UI action, not an expressive prompt. That is correct per voice.md.

One spelling note: the codebase uses UK English throughout (voice.md is explicit). The Maestro flow comment uses "Pre-condition" which is fine, but "summarise" and similar words should be checked in any user-visible strings added in this PR. The visible strings in this diff ("Show context", "Hide context") are not affected by UK/US spelling, so no blocking issue there.

### Interaction

The Show/Hide toggle is inline beneath the entry row, which fits the established expand/collapse pattern already used for entry detail (expandedIds). No unexpected interruptions or confirmations. The context clears automatically when the filter changes — this is the right behaviour and prevents stale context from confusing the user.

The empty-result dead end (described above) is the only interaction gap.

### Side Effects

The diff also introduces the Focus Quick-Log screen (`app/(tabs)/(tend)/focus/[id]/index.tsx`) and its tests. This is not in scope for issue #137 or user story #118. It does not appear to break any existing flow, but it is worth confirming that this was intentional bundling and that the Quick-Log screen has its own issue/story tracking it — otherwise product coverage for that screen is invisible in the milestone.

### Summary of Concerns

1. **Empty context result is a silent dead end.** Tapping "Show context" when no surrounding entries exist gives zero feedback. Needs a quiet inline message — this is a UX gap, not a blocking data-safety issue.
2. **Out-of-scope file bundled into this PR.** The Quick-Log screen is a substantial addition that does not appear to be part of #137 or #118. Confirm this is tracked and intentional.

sfrankle (commented):
## Code Quality Review

**Verdict:** CONCERNS

### React Native / Expo

**handleShowContext stale closure on contextMap (minor):** The `handleShowContext` callback lists `contextMap` in its `useCallback` dependency array (trace.tsx, renderItem deps). This is correct for correctness — the guard `if (contextMap.has(entry.id)) return` needs the current map — but it means every context fetch causes `handleShowContext` to be recreated, which causes `renderItem` to be recreated, which causes all visible rows to re-render. For a long Trace list this could be noticeable. A common pattern to avoid this is to read the map inside the setter callback (functional update) rather than closing over it, or using a `useRef` shadow for the guard. Not a block, but worth addressing.

**Hardcoded pixel values in styles:** `contextButton` uses `paddingVertical: 6` and `contextRow` uses `paddingVertical: 10`. The design system provides `spacing.tight` and `spacing.micro`. Every other style block in this file uses spacing tokens. These two values should go through the token.

### TypeScript

**`as never` type assertion in quick-log screen:** `router.replace(`/focus/${focusId}/edit` as never)` at line ~354 of `app/(tabs)/(tend)/focus/[id]/index.tsx`. This is a real type assertion hiding the fact that Expo Router's typed routes don't know about this path at compile time. The existing codebase uses `as never` for Expo Router push/replace calls elsewhere as a known workaround, so this is consistent — but it is worth a comment noting that the route is typed and the assertion can be removed once all routes are registered in Expo Router's typed manifest.

**Double cast `as unknown as Db`:** Used in both `handleSubmit` (quick-log) and `handleShowContext` (trace). This is consistent with the existing codebase pattern for bridging expo-sqlite and the `Db` interface in tests, so it is not a new concern — but both sites should carry the same inline comment that exists on the pattern in other screens explaining why the cast is needed.

### Haven Patterns

**Missing decisions.md entry for the "Show context" feature:** The PR adds three good decisions.md entries (saveEntryBatch, quick-log LogFormShell exemption, isPhysical). However, the core Trace-screen decision — `contextMap` state, the ±2-hour window fetched on demand via `getContextEntries`, and the clear-on-filter-change behaviour — has no entry. Per Haven's pattern, decisions with architectural implications (new state shape on a primary screen, new data-fetching approach keyed to UX interaction) require a row. This should be added before merge.

### Tests

**Plan critic concern resolved — muted style assertion present:** The test `'context entry rows have muted style (opacity 0.7)'` in `src/__tests__/screens/TraceScreen.test.tsx` directly asserts `flatStyle.opacity === 0.7` on `testID="context-row-99"`. The plan critic concern is addressed.

**TDD order confirmed:** Commit history shows failing tests (`288ad59`) precede implementation commits (`7b68203`, `27310fc`). No concern.

**Test coverage is solid:** Empty window, exclusion of focal entry, label collapsing, filter-change clearing, and hide behaviour are all covered. No gaps found.

### General

No dead code, no console.log, no commented-out blocks. Changelog is correct (#154 entry describes this PR's feature). The `collapseTraceRows` extraction is clean and the JSDoc on `getContextEntries` explaining why the caller owns time arithmetic is a good practice.

sfrankle (commented):
## Product Vision Review

**Verdict:** PASS

### Findings

**Empty State (targeted re-check)**
The previously flagged silent dead end is resolved. When contextEntries is defined but empty, line 141 renders "Nothing else logged around this time." with the expected testID. The copy is warm and neutral — no negative framing, no implication of failure. This is consistent with Haven's empty-state principles.

sfrankle (commented):
## Code Quality Review

**Verdict:** PASS

### Findings

#### Haven Patterns
Fixed: decisions.md entry for the Trace context pattern is present (2026-04-16 row). It covers contextMap, on-demand fetch, cleared-on-filter and cleared-on-tab-nav, and inline rendering rationale.

#### React Native / Hooks
Fixed: handleShowContext stale closure resolved. The guard against double-fetch is now inside the setContextMap functional updater (prev.has(entry.id)), so it always reads current state rather than a closed-over stale reference. Dependency array is correctly empty.

#### Previously acknowledged — not reviewed
Hardcoded paddingVertical values (6 and 10) were excluded from this targeted re-review per the orchestrator's instructions.

Code quality looks good.

PR #155, opened 2026-04-17
Physical: log severity 0 — symptom absent today

### LINKED ISSUE #140: Physical: log severity 0 — symptom absent today
## Problem

The current severity scale starts at 1, which means a tracked symptom can only be logged when it is present. For over-time tracking — e.g. hand pain across weeks — the absence of the symptom is equally meaningful data. Without severity 0, a pain chart has no zero values and isn't a real chart.

## Desired behaviour

When logging a Physical entry, severity 0 is a valid selection meaning "this symptom is not occurring today." It should be loggable so that Weave can plot the full picture over time including absence.

## Notes

- Closely related to #139 (symptom + location model rethink) — if tracking `Pain + Wrist` over time, severity 0 means "no wrist pain today". These two features should be designed together.
- The use case is primarily Focus-driven: a user monitoring a named concern wants to log check-ins including zero days, not just flare days.
- UI implication: the severity row (currently 1–5) needs to accommodate 0 without looking like an error state or empty state.

Contributes to #111

sfrankle:
## Summary

- Extends the Physical severity scale from 1–5 to 0–5, where 0 means the tracked symptom is absent today
- Adds a dedicated "0" button to SeverityRow so users can explicitly log zero-severity check-ins — filling in the missing data points that make over-time charts meaningful
- Extracts `formatChipLabel` into a shared utility and renders "(absent)" for severity-0 chips in Trace so logged absences are readable in history

Closes #140
Contributes to #111

## What to review

- `SeverityRow` — layout with the added 0 button; make sure it doesn't read as an error/empty state
- `formatChipLabel` utility — extracted from existing inline logic; verify all callers still behave correctly
- Trace chip rendering for severity 0 — the "(absent)" label is the only place this semantic is surfaced to the user

## Testing

- Updated unit tests for `formatChipLabel` covering severity 0
- Manual: select a Physical label, tap 0 in severity row, save; confirm chip shows "(absent)" in Trace

## Checklist
- [x] Acceptance criteria met
- [x] CI passing
- [x] docs/changelog.md updated

### DIFF
--- a/app/(tabs)/(tend)/log/physical.tsx
+++ b/app/(tabs)/(tend)/log/physical.tsx
@@ -24,48 +24,15 @@ import {
 } from '@/lib/db/queries';
 import { getDb } from '@/lib/db/database';
 import { nowLocalIso } from '@/lib/utils/timestamp';
+import { formatChipLabel } from '@/lib/utils/physicalChipLabel';
+import type { EnergyChip, StateChip, PhysicalChip } from '@/lib/utils/physicalChipLabel';
 import { colors, lineHeight, spacing, typeScale } from '@/constants/theme';
 import { colorForPhysicalLabel } from '@/constants/chipColors';
 import { logScreenStyles } from '@/constants/sharedStyles';
 import type { Db } from '@/lib/db/queries';
 import type { PhysicalStateLabel } from '@/lib/db/query-types';
 
-// ─── Types ─────────────────────────────────────────────────────────────────
-
-type EnergyChip = {
-  kind: 'energy';
-  id: 'energy';
-  value: number; // 1-5
-};
-
-type StateChip = {
-  kind: 'state';
-  id: number; // labelId
-  labelName: string;
-  parentName: string | null;
-  severity: number | null; // 1-5 or null
-};
-
-type PhysicalChip = EnergyChip | StateChip;
-
 // ─── Helpers ───────────────────────────────────────────────────────────────
-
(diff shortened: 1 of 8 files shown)


sfrankle (commented):
## Simplify Pass

Two fixes applied after the three-agent review:

1. **Reuse**: `physicalChipLabel.ts` was duplicating `WHOLE_BODY_NAMES` and the area-prefix check already exported as `shouldShowAreaPrefix` from `traceUtils.ts`. Now imports and calls the shared helper.
2. **Robustness**: `summariseEntry` in `traceUtils.ts` used a falsy-ordering trick where the `=== 0` early-return was needed to keep `0` out of the `!= null` branch. Made explicit: `numericValue !== null && numericValue !== 0` so the intent survives future edits.

All 35 affected tests pass.

sfrankle (commented):
## Code Quality Review — traceUtils.ts follow-up

**Verdict:** PASS

Verified the fix at lines 65–68 of `src/lib/utils/traceUtils.ts`. The redundant `&& numericValue !== 0` has been removed. The early return on `numericValue === 0` now correctly short-circuits before the ternary, leaving clean, non-contradictory logic:

```ts
if (numericValue === 0) return `Felt ${stateName} (absent)`;
return numericValue !== null
  ? `Felt ${stateName} (${numericValue}/5)`
  : `Felt ${stateName}`;
```

Dead code concern is resolved.

Issue #157, opened 2026-05-22
Trace: group Routine rows in filtered view when Focus filter is active


sfrankle:
## Background

Issue #156 (Focus user feedback) deferred this item — the Routines feature (#117) is not yet built, so the grouping behaviour could not be implemented.

## Acceptance Criterion

From issue #117 AC: when a Focus filter is active on the Trace screen, entries that belong to the same Routine should be visually grouped (e.g. collapsed under a routine header row) to reduce noise and make patterns more scannable.

## When to implement

After Routine logging is built and entries have a `routine_id` or equivalent link. At that point, the Trace filtered view should group consecutive Routine-linked entries under a single collapsible header.

## References
- Parent issue: #156 (Focus user feedback — item 7 deferred)
- Routines feature: #117

sfrankle:
Delivered by #182 (Routine grouping in Trace, multi-filter support).

Verified against this issue's acceptance criterion:

- `buildTraceItems` (`src/lib/utils/traceUtils.ts`) collapses entries sharing a `routine_completion_id` into a single group row, emitted at the position of its newest member so surrounding chronology is preserved.
- `useTraceEntries` runs that grouping over the focus-filtered entry stream (`getEntriesForTrace(db, { focusIds })`), so grouping is active whenever a Focus filter is.
- `app/(tabs)/trace.tsx` renders each group via `RoutineGroupRow` with `expanded` / `onToggle` — the collapsible header row this issue described.

Two things went beyond the original scope: grouping applies unfiltered as well as filtered, and `matchedIds` mutes group members that do not match the active filter rather than hiding them.

Closing as already implemented — no separate work needed.

PR #158, opened 2026-05-22
Focus: user feedback polish (#156)

### LINKED ISSUE #156: Focus: user feedback
- [ ] when creating new Focus from inside area (Nourish, Replenish, etc) keyboard blocks textbox with for Focus name
   - [ ] also cannot associate labels from this view, which is fine, but contradicts `Both paths use the same screen: name field (required) + optional tracked labels picker` from https://github.com/sfrankle/haven-app/issues/114
- [ ] Tend dashboard -> "Jaw" (Focus) -> selecting severity for label does not include 0 (this was recently added.)
   - there should be 1 source of truth for the rules. clearly Attune severity rules are in multiple places since 0 is not available here but is on the primary Attune page
- [ ] I want to be able to delete or hide Focus on the Trace screen
- [ ] in Attune page -> select sensation, set severity, add Focus, cannot scroll to see save button (hidden under bottom nav)
- [ ] Unveil - Focus menu needs padding / margin on left side
- [ ] When I navigate to Trace, prior filters are still in place. This is ok. If I'm in Trace and I click Trace again, it should reset the filters. (Aka remove selected Focus)
- [ ] `Grouped Routine rows remain grouped when a filter is active — the group is shown if any entry in the group matches the Focus` from https://github.com/sfrankle/haven-app/issues/117 is not complete, as Routines have not been completed. Make sure this is tracked somewhere else.
- [ ] Show context: I want to have an indicator where the record is in relation to the context. Also, now I'm seeing duplicate records. see comment for details.
- [ ] Clicking into notes -> keyboard is over Notes field so I can't see what I'm typing - not good. see comment for image
- [ ] Settings -> Focus. There needs to be more visual difference. Option: put focus settings in a nested page. Option: better use of H1 vs H2 etc and maybe a line section divider. see image in comment.

sfrankle:
## Summary

- Addresses 9 user-reported issues from Focus milestone QA: keyboard avoidance, severity gaps, Trace UX, scroll safety, and visual hierarchy.
- Each fix is self-contained; the one deferred item (Routines grouping) is tracked separately as #157.
- No schema changes; all fixes are UI/interaction layer.

Closes #156

## What to review

- `FocusDropdown` keyboard avoidance — modal now lifts above the keyboard so the name field stays visible.
- Severity 0 in Focus Quick-Log — unified the severity rules so zero ("absent") is available from the Tend quick-log path, matching the primary Attune screen.
- Long-press Focus pill on Trace — long-pressing a Focus pill now routes to the Edit/Archive screen.
- Attune physical screen — inner `View` replaced with `ScrollView` so the save button is always reachable when the bottom nav is visible.
- `FocusDropdown` left padding on emotion screens — `style` prop threads the correct padding into the dropdown on Unveil screens.
- Trace tab re-tap resets filter — tapping the already-active Trace tab now clears the active Focus filter.
- Show Context view mode — context view now replaces the filtered list (focal entry gets a glow border, non-focus entries are muted, dismiss returns to filter) and the duplicate-record bug is resolved.
- Notes keyboard avoidance — `KeyboardAvoidingView` added to emotion tier2 and tier3 so the Notes field stays visible while typing.
- Settings visual hierarchy — section divider and increased heading weight distinguish Focus settings from adjacent content.

## Testing

Each item was tested manually against the repro steps in the issue comments. No new Maestro flows were added (all changes are within existing screens with existing flows). No migration involved.

## Checklist
- [x] Acceptance criteria met
- [x] CI passing


sfrankle (commented):
## Simplify Pass

Four issues fixed:

- **KAV copy-paste (tier2/tier3):** Added `avoidKeyboard` prop to `Screen` component — tier2 and tier3 now use `<Screen showBack avoidKeyboard>` instead of wrapping content manually in `KeyboardAvoidingView`.
- **Redundant state:** Merged `contextFocalEntry` + `contextEntries` into a single `contextState?: { focalEntry, entries }` object. All resets are now atomic — one `setContextState(undefined)` instead of two separate calls.
- **Duplicate icon block:** Extracted `EntryIcon` helper in `trace.tsx` — the icon/placeholder pattern was duplicated in both `EntryRow` and `ContextViewRow`.
- **Inline renderItem:** `ContextView`'s `FlatList` renderItem was an inline arrow function recreated on every render; extracted to `useCallback`.

sfrankle (commented):
## Code Quality Review

**Verdict:** PASS

### Re-review scope

This is a targeted re-review of the two fixes applied in commit 33d19df. The broader PR was previously reviewed; only the two flagged issues are re-assessed here.

---

### Fix 1: docs/decisions.md — stale contextMap row replaced

Confirmed. The 2026-04-16 `contextMap` row is gone. Two new rows (both dated 2026-05-22) appear at the end of the table, as required by the append-only convention:

- **Context view mode redesign** — accurately describes the `contextState` single-object pattern, the FlatList replacement, the muted/full-opacity rendering split, and the cleared-on-filter/tab-press/useFocusEffect lifecycle.
- **tabPress listener pattern** — documents the `useNavigation().addListener('tabPress')` workaround, the `isFocused()` guard rationale, and the `'tabPress' as any` cast with a pointer to the eslint-disable comment at the call site. The escape hatch (migrate when Expo Router exposes a first-party API) is present.

Both entries meet the decisions.md standard: date, decision, rationale, detail, and guidance for future callers. No issues.

---

### Fix 2: app/(tabs)/__tests__/settings.test.tsx — divider test corrected

Confirmed. The test at line 36 now reads:

> `'shows a section divider between Privacy and Focus (always rendered)'`

`mockGetFocuses.mockResolvedValue([])` is used (empty focuses), which correctly exercises the unconditional render path. The previous version used `[ACTIVE_FOCUS]`, which asserted the right thing for the wrong reason — it would have passed even if the divider were conditionally gated on focuses existing.

The corrected test now falsifies the conditional-render hypothesis: with zero focuses the divider must still appear, which is exactly what the implementation guarantees.

No other tests in the file were disturbed. The rest of the describe blocks (`'hides Focus section when there are no focuses'`, error rollback, optimistic hide) remain correct.

---

### No new concerns introduced

The commit touches exactly two files and makes no logic changes. Code quality looks good.

Issue #163, opened 2026-05-22
Tend dashboard — Routines section


sfrankle:
## Summary
Add a Routines section to the Tend dashboard, above the Focus pills row, showing due-now cards and a collapsed disclosure row for completed/not-yet-due Routines.

## Acceptance criteria
- [ ] `RoutineSection` component inserted above `FocusRow` in `app/(tabs)/(tend)/index.tsx`
- [ ] When no active Routines exist, section shows only a `+ Add Routine` entry (navigates to create screen)
- [ ] Archived Routines never appear
- [ ] Due-now cards (Routine due in current time block, not yet completed this block): show Routine name, configured time blocks, completion count for today (e.g. "1 of 2 · Morning done"); tapping opens complete-Routine flow
- [ ] Collapsed disclosure row below active cards contains: completed Routines + not-yet-due Routines; label reflects contents (e.g. "Completed · Later"); tap expands; completed Routines show completion count; user can tap any expanded Routine to complete early
- [ ] No "overdue" or "missed" language — a Routine not yet done for a past block is simply still accessible, never penalised
- [ ] Tests: unit tests for RoutineSection covering empty state, due cards, disclosure row expand/collapse

## Notes
Completion state derived from `useRoutineCompletionStates` hook (#160). No extra completion state is stored — it is computed from `routine_completion` + current time block + today's date.

Blocked by: #160

sfrankle:
## Implementation Plan

**Approach:** Add a `RoutineSection` above `FocusRow` on Tend. Completion classification comes from `useRoutineCompletionStates` (#160) unchanged; a new display-only batched read (`getRoutineDayProgress`) supplies the "1 of 2 · Morning done" counts, which the three-value state enum cannot express. All grouping and copy logic lives in a pure, fully-tested `routine-dashboard.ts` so the spec's worked examples map one-to-one onto unit tests. **No schema change** — everything is derived from existing `routine_completion` rows.

**What will change:**
- New: `src/lib/utils/routine-dashboard.ts` (grouping + copy formatters), `src/hooks/useRoutineDayProgress.ts`, `src/components/RoutineSection.tsx`, `src/components/RoutineCard.tsx`
- Modified: `src/lib/db/queries.ts` + `query-types.ts` (add `getRoutineDayProgress`, read-only), `src/hooks/useRoutineCompletionStates.ts` (optional `refreshKey` param — see risks), `app/(tabs)/(tend)/index.tsx` (mount the section), hooks/components barrel exports
[...]

Issue #123, opened 2026-04-08
Create and manage a Routine


sfrankle:
## Story
As a user, I want to create, edit, and archive Routines so that I can define what I do repeatedly, refine it as my needs change, and retire it when it's no longer relevant.

### Creating (new)
- [ ] Accessible from "+ Add Routine" in the Routines section on Tend
- [ ] Accessible from Settings → Routines management screen
- [ ] Creation flow: name → associated Focus (optional) → time blocks (multi-select) → items → save
- [ ] Frequency note field (optional, informational only — e.g. "3x daily as prescribed")

### Routine items
- [ ] User adds items one at a time; each becomes a card in the ordered list
- [ ] Per item: name (required), entry type, labels (required for most types; Hydration and Sleep are numeric-only), prescribed detail (optional free text, pre-populates `entry.notes` at completion), instruction note (optional, longer description)
- [ ] Items can be reordered

### Editing (from view state)
- [ ] Edit mode accessible from the Routine card on Tend (long press or settings icon)
- [ ] Edit mode accessible from Settings → Routines management screen → select Routine
- [ ] Both paths use the same edit screen; all fields are editable
- [ ] Editing prescribed detail (e.g. reps change from 10 to 12) takes effect immediately for future completions
- [ ] Past log entries are never modified — they captured what the user actually did

### Archiving
- [ ] Archive option available in edit mode
- [ ] Archived Routine disappears from Tend dashboard immediately
- [ ] Archived Routines listed in Settings, with option to unarchive
- [ ] Past entries created by this Routine remain unchanged and visible in Trace

## Notes
Frequency is informational reference only — Haven does not cross-check frequency against time blocks or generate warnings if they differ.

PR #174, opened 2026-06-05
Settings — Routines management section

### LINKED ISSUE #162: Settings — Routines management section
## Summary
Routines management section in Settings, mirroring the existing Focus section pattern.

## Acceptance criteria
- [ ] Settings screen gains a Routines section below the Focus section
- [ ] Active Routines listed with name; tap navigates to edit screen
- [ ] Archived Routines listed below active; unarchive button restores immediately (optimistic update, same pattern as archived Focus unarchive)
- [ ] When no Routines exist, section shows a `+ Add Routine` entry
- [ ] Tests: unit tests covering active/archived rendering and unarchive interaction

## Notes
Follow the `SettingsFocusSection` pattern in `app/(tabs)/settings.tsx` for list structure, optimistic unarchive state, and style.

Blocked by: #160

sfrankle:
## Summary
- Adds a Routines management section to the Settings screen, directly below the Focus section.
- Mirrors the existing `SettingsFocusSection` pattern: active Routines list with an edit affordance, an Archived group with optimistic unarchive, and a persistent `+ Add Routine` entry.
- Lets users edit a Routine from Settings and restore an archived one without leaving the screen.

Closes #162
Contributes to #123

## What to review
- `app/(tabs)/settings.tsx` — the new `SettingsRoutinesSection` component. Worth a look: the optimistic unarchive uses a local `locallyUnarchived` Set (same approach as the archived Focus unarchive) so the row disappears immediately while the write is in flight, and a failed write leaves the row visible for retry.
- `+ Add Routine` renders persistently at the bottom of the section (not gated on the empty state) so users with existing Routines can always create another — Settings is the canonical create-anytime surface while the Tend entry point is deferred to #163.
- Section ordering and divider placement in `SettingsScreen`.

## Testing
- Added unit tests in `app/(tabs)/__tests__/settings.test.tsx` covering active and archived Routine rendering, the persistent `+ Add Routine` entry, edit navigation, and the unarchive interaction.
- Added a Maestro flow `maestro/flows/settings/settings-routines.yaml` asserting the Routines section and `+ Add Routine` entry are visible on a fresh install.
- Updated `maestro/flows/tend/routine-create-edit.yaml` notes now that a live Settings entry point exists.
- `npx tsc --noEmit`, `npx eslint`, and `npm test` run locally.

## Checklist
- [x] Acceptance criteria met
- [x] CI passing
- [x] \`docs/changelog.md\` updated

### DIFF
--- a/app/(tabs)/__tests__/settings.test.tsx
+++ b/app/(tabs)/__tests__/settings.test.tsx
@@ -172,18 +172,20 @@ describe('SettingsScreen — Routines section', () => {
     expect(mockPush).toHaveBeenCalledWith('/routine/create');
   });
 
-  it('hides + Add Routine when active routines exist', async () => {
+  it('keeps + Add Routine visible alongside active routines', async () => {
+    // Settings is the canonical create-anytime surface, so + Add Routine is
+    // always available — not gated on the empty state.
     mockGetRoutines.mockResolvedValue([ACTIVE_ROUTINE]);
-    const { queryByTestId, getByTestId } = render(<SettingsScreen />);
+    const { getByTestId } = render(<SettingsScreen />);
     await waitFor(() => getByTestId(`settings-routine-row-${ACTIVE_ROUTINE.id}`));
-    expect(queryByTestId('settings-routines-add-row')).toBeNull();
+    expect(getByTestId('settings-routines-add-row')).toBeTruthy();
   });
 
-  it('hides + Add Routine when only archived routines exist', async () => {
+  it('keeps + Add Routine visible alongside archived routines', async () => {
     mockGetRoutines.mockResolvedValue([ARCHIVED_ROUTINE]);
-    const { queryByTestId, getByTestId } = render(<SettingsScreen />);
+    const { getByTestId } = render(<SettingsScreen />);
     await waitFor(() => getByTestId(`settings-routine-archived-row-${ARCHIVED_ROUTINE.id}`));
-    expect(queryByTestId('settings-routines-add-row')).toBeNull();
+    expect(getByTestId('settings-routines-add-row')).toBeTruthy();
   });
 
   it('shows active routine rows with a gear icon', async () => {
@@ -228,9 +230,9 @@ describe('SettingsScreen — Routines section', () => {
     });
   });
 
-  it('does not show + Add Routine after optimistically unarchiving the only routine', async () => {
-    // The hook does not refetch, so the unarchived routine still counts as an
-    // existing routine — + Add Routine must stay hidden (allRoutines.length > 0).
+  it('keeps + Add Routine visible after optimistically unarchiving the only routine', async () => {
+    // + Add Routine is always present, so it stays visible through the
+    // optimistic unarchive (the archived row disappears, the add row does not).
     mockGetRoutines.mockResolvedValue([ARCHIVED_ROUTINE]);
     const { getByTestId, queryByTestId } = render(<SettingsScreen />);
     await waitFor(() => getByTestId(`settings-routine-unarchive-${ARCHIVED_ROUTINE.id}`));
@@ -242,7 +244,7 @@ describe('SettingsScreen — Routines section', () => {
     await waitFor(() => {
       expect(queryByTestId(`settings-routine-archived-row-${ARCHIVED_ROUTINE.id}`)).toBeNull();
     });
-    expect(queryByTestId('settings-routines-add-row')).toBeNull();
+    expect(getByTestId('settings-routines-add-row')).toBeTruthy();
   });
 
   it('shows the archived row again when setRoutineArchived throws', async () => {
--- a/app/(tabs)/__tests__/settings.test.tsx
+++ b/app/(tabs)/__tests__/settings.test.tsx
@@ -15,9 +15,13 @@ jest.mock('@/lib/db/database', () => ({
 
 const mockSetFocusArchived = jest.fn();
 const mockGetFocuses = jest.fn();
+const mockSetRoutineArchived = jest.fn();
+const mockGetRoutines = jest.fn();
 jest.mock('@/lib/db/queries', () => ({
   getFocuses: (...args: unknown[]) => mockGetFocuses(...args),
   setFocusArchived: (...args: unknown[]) => mockSetFocusArchived(...args),
+  getRoutines: (...args: unknown[]) => mockGetRoutines(...args),
+  setRoutineArchived: (...args: unknown[]) => mockSetRoutineArchived(...args),
 }));
 
 // ─── fixtures ─────────────────────────────────────────────────────────────────
@@ -25,12 +29,24 @@ jest.mock('@/lib/db/queries', () => ({
 const ACTIVE_FOCUS = { id: 1, name: 'Energy', description: null, archived: false, sortOrder: 0, createdAt: '' };
 const ARCHIVED_FOCUS = { id: 2, name: 'Old Focus', description: null, archived: true, sortOrder: 1, createdAt: '' };
 
+const ACTIVE_ROUTINE = {
+  id: 10, name: 'Morning Reset', associatedFocusId: null,
+  frequencyNote: null, sortOrder: 0, archived: false,
+  createdAt: '', updatedAt: '', timeBlocks: [],
+};
+const ARCHIVED_ROUTINE = {
+  id: 11, name: 'Old Routine', associatedFocusId: null,
+  frequencyNote: null, sortOrder: 1, archived: true,
+  createdAt: '', updatedAt: '', timeBlocks: [],
+};
+
 // eslint-disable-next-line import/first
 import SettingsScreen from '../settings';
 
 describe('SettingsScreen — visual hierarchy', () => {
   beforeEach(() => {
     jest.clearAllMocks();
+    mockGetRoutines.mockResolvedValue([]);
   });
 
   it('shows a section divider between Privacy and Focus (always rendered)', async () => {
@@ -54,6 +70,7 @@ describe('SettingsScreen — Focus section', () => {
   beforeEach(() => {
     jest.clearAllMocks();
     mockSetFocusArchived.mockResolvedValue(undefined);
+    mockGetRoutines.mockResolvedValue([]);
   });
 
   it('hides Focus section when there are no focuses', async () => {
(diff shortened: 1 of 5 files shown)


sfrankle (commented):
## Simplify Pass

Ran a 4-angle cleanup review (reuse / simplification / efficiency / altitude) over the diff.

**Fixed:**
- Collapsed the new `addRoutineText` style into the pre-existing identical `unarchiveText` style — both were `labelMedium` + `colors.interactive`. Renamed the shared style to `interactiveLabel` and pointed all three call sites (Routine add button + both Unarchive buttons) at it. Removes 6 lines of duplication and a future theme-drift hazard.

**Reviewed and intentionally skipped:**
- **Extracting a generic `ArchivableListSection`** from the duplicated Focus/Routines section components — 3 of 4 agents judged extraction premature at only 2 instances, and the sections already diverge (Focus hides when empty; Routines shows `+ Add Routine`). Revisit at a 3rd consumer.
- **`useMemo` on the active/archived filters** — the established `SettingsFocusSection` in the same file doesn't memoize; adding it only to Routines would break symmetry for negligible gain on a tiny list.
- **Lifting optimistic-unarchive into the hook** — defensible as-is (self-heals via `useFocusEffect` refetch); worth a follow-up only once 3+ entities share the pattern.

Verification: `npx tsc --noEmit` clean, `eslint` clean, Routines test suite 12/12 green.

sfrankle (commented):
## Safety Review

**Verdict:** PASS

### Privacy: ✅ No new fetch/axios/HTTP client, analytics, telemetry, or crash reporting. The only `expo-*` reference is the pre-existing `expo-router` import (navigation only, no off-device access). No new network permissions or external-URL libraries.

### Data Safety: ✅ No schema migration needed — `routine.archived` already exists, and no migration/schema files are touched. The optimistic unarchive is safe: `setRoutineArchived(db, id, false)` is awaited, and the local `locallyUnarchived` Set is updated only on success. On failure the catch leaves the Set unchanged, so the archived row reappears for retry — no silent data loss. The optimistic hide is a pure UI filter, never a DB write, so the database write remains the single source of truth. Failure path is covered by the test 'shows the archived row again when setRoutineArchived throws'. No `INSERT OR REPLACE`/plain `INSERT`, no column/table drops.

### Judgmental Language: ✅ No matches for good/bad/great/failed/missed/behind/streak/score/rank/incomplete in user-facing strings. New copy ('Routines', '+ Add Routine', 'Unarchive', 'Archived', 'Edit ${name}') is neutral and non-pressuring.

### Workflow Artifacts: ✅ PR body has `Closes #162` (technical task) and `Contributes to #123` (user story). `docs/changelog.md` has exactly 1 new row, correctly keyed to PR #174 (not the issue number), placed at the top.

All safety checks passed.

sfrankle (commented):
## Product Vision Review

**Verdict:** BLOCK

### Findings

#### User Story Fulfillment (BLOCK)
The user story (#123) acceptance criteria states, unconditionally:
> Accessible from Settings → Routines management screen

The implementation only satisfies this at zero-state. `+ Add Routine` renders solely when `allRoutines.length === 0`. Once a user has **any** routine — active or archived — the create entry point disappears from Settings.

I confirmed by grep that Settings is currently the **only** in-app entry point to `/routine/create`. The Tend `+ Add Routine` entry point is explicitly deferred to a separate issue (per the note in `maestro/flows/tend/routine-create-edit.yaml`).

**Net effect: a user who has created at least one routine cannot create another routine anywhere in the app.** This is a core action made unreachable.

#### Interaction — dead end (BLOCK)
The PR note reasons that 'the archived list provides the path back otherwise.' This does not hold: unarchiving restores a *previously created* routine — it is not a path to create a *new* one. So the states are:
- 0 routines → `+ Add Routine` visible (create works)
- ≥1 active routine, no way to add a second
- only archived routines → no create path, and unarchive only resurrects old ones

All non-empty states are dead ends for creation.

**Minimal fix:** surface `+ Add Routine` whenever the Routines section renders (e.g. at the bottom of the list), not gated on `allRoutines.length === 0`.

#### Side Effects (CONCERNS, non-blocking)
On a fresh install `SettingsFocusSection` returns `null`, but both `settings-section-divider` and `settings-routines-divider` still render (the test 'always renders the second section divider, even with no routines' confirms this). Result: two stacked dividers with no content between Privacy and Routines. Minor visual artefact; worth tidying but does not affect the verdict.

### What is clean
- **Tone & copy:** 'Routines', '+ Add Routine', 'Unarchive', 'Archived' are all neutral and on-voice. No scores, streaks, ranking, or pressure language. Aligns with `docs/design/voice.md`.
- **Empty state:** quietly inviting (`+ Add Routine`), not negative or guilt-inducing — correct per Haven's empty-state rules.
- **Optimistic unarchive:** matches the established Focus pattern; failed write leaves the row visible for retry — neutral, no alarm.
- **Edit affordance:** gear icon mirrors the Focus section, so the flow feels consistent with the rest of Settings.

The implementation is well-aligned in tone and pattern; the block is targeted at the unreachable create action, which both breaks the user story AC and leaves a real interaction dead end.

sfrankle (commented):
## Code Quality Review

**Verdict:** CONCERNS

The implementation is a clean, faithful mirror of `SettingsFocusSection` and the three plan-critic concerns are all handled correctly (see Verified below). One actionable issue and one cosmetic note.

### RN / Expo

**[Concern] Load-time flash of the wrong affordance.** `SettingsRoutinesSection` destructures only `routines` from `useRoutines` (settings.tsx:112) and ignores `loading`. `allRoutines` starts as `[]`, and `+ Add Routine` is gated purely on `allRoutines.length === 0` (settings.tsx:140). `useRoutines` reloads on every `useFocusEffect` (gains focus), so during each async load window the section renders `+ Add Routine` first, then swaps to the routine rows once `getRoutines` resolves. For any user who actually has routines, that is a visible flash of the wrong content every time they open Settings.

This is the one place the mirror diverges meaningfully from the source: `SettingsFocusSection` does `if (allFocuses.length === 0) return null` (settings.tsx:57), so it renders nothing during load and never shows wrong content. Routines cannot return null by design (the add row must stay reachable — correct, per decision row 34's reasoning), but the add row should be gated on `!loading && allRoutines.length === 0` to suppress the flash.

Note the unit tests will not catch this: the `useFocusEffect: (cb) => cb()` mock plus an already-resolved `getRoutines` collapse the load window to zero. A test that asserts `+ Add Routine` is absent while the promise is still pending would cover it.

### Haven Patterns

No new `docs/decisions.md` entry is required — this is not a new pattern. Row 37 establishes Settings as the canonical management surface (gear → edit, unarchive for archived), and row 47 explicitly anticipates 'a future Settings → Routines unarchive flow' as the escape hatch for no-confirm archive. This PR implements that already-decided behavior.

The `(await getDb()) as unknown as Db` double cast is consistent with the existing `SettingsFocusSection` and `useRoutines` — not a new finding.

### General

**[Minor / cosmetic] Stacked dividers in the no-focus state.** When `allFocuses.length === 0`, `SettingsFocusSection` returns null, so the render is `divider → (nothing) → divider → Routines` (settings.tsx:220–226): two dividers with no content between them. A single dangling divider after Privacy already existed pre-PR in this state; this PR introduces the doubling. Non-blocking, but worth a guard if the empty-Focus state is reachable in practice.

### Tests

Good coverage: heading-always-visible, add-row visibility across zero/active/archived, edit and create navigation, optimistic unarchive (success, persistence after the only routine is unarchived, and DB-error rollback). Tests assert behavior via testIDs, names describe behavior. Gap: no test for the load-window flash above.

### Verified (plan-critic concerns)
- Fixture/testID id mismatch — resolved; `${ACTIVE_ROUTINE.id}` / `${ARCHIVED_ROUTINE.id}` constants used throughout.
- `+ Add Routine` shows only when `allRoutines.length === 0` (zero routines at all) — correct; covered by 'hides + Add Routine when only archived routines exist' and the post-unarchive test.
- `/routine/create` resolves from Settings — correct; `(tend)` is a route group (no URL segment), same proven mechanism `SettingsFocusSection` uses for `/focus/${id}/edit`.

sfrankle (commented):
## Code Quality Review (follow-up)

**Verdict:** PASS

### Prior concern — resolved
The empty→populated flash of `+ Add Routine` is gone. Commit 41f957d makes the add row a persistent, unconditionally-rendered row at the bottom of the section (`settings-routines-add-row`, settings.tsx:184-195) — it no longer reads emptiness at all, so there's no data-dependent transition that can flash. The product decision (Settings is the canonical create-anytime surface) is documented in an inline comment and locked in by tests asserting the row stays visible alongside active routines, alongside archived routines, and through an optimistic unarchive of the only routine.

### Findings

**RN/Expo:** Hooks clean. `useRoutines({ includeArchived: true })`, optimistic-unarchive state, and the `as unknown as Db` cast all mirror the existing `SettingsFocusSection` pattern — consistent, no new conventions introduced. The `loading` flag remains unused, but it is now moot for the flagged concern since the add row no longer depends on it.

**TypeScript:** No new `any`. `Routine` typed via `query-types`. Cast matches existing pattern.

**Tests:** Strong, behavior-focused coverage — empty / active / archived / navigation / optimistic unarchive / error rollback, plus explicit add-row-persistence assertions. Suite runs green (22 passed).

**General:** No dead code, no console.log, no commented-out blocks. The shared `unarchiveText`→`interactiveLabel` style rename is a reasonable cleanup.

### Non-blocking notes
- Active routine *rows* (settings.tsx:140-157) still gate on `allRoutines`, so they could theoretically blink if `useRoutines` clears to `[]` during a focus reload. This was never the flagged concern (which was the add row specifically) and is the identical shared pattern the shipped Focus section already uses. No change requested.
- The fresh-install double-divider remains, acknowledged as intended and matching existing Privacy→Focus behavior. Settled.

Code quality looks good.

sfrankle (commented):
## Product Vision Review (re-review)

**Verdict:** PASS

### Blocking issue resolved
The prior BLOCK is fixed. `+ Add Routine` is now a persistent row (`settings-routines-add-row`) rendered unconditionally at the bottom of the Routines section, independent of routine count. A user with existing routines can create another one — Settings is the canonical create-anytime management surface. Tests assert it stays visible alongside active routines, archived routines, and through an optimistic unarchive. Verdict moves to PASS.

### Findings

**User Story Fulfillment** — AC for #162 met. Note the implementation is *better* than the literal AC wording (which said `+ Add Routine` appears 'when no Routines exist'): making it always-present matches the #123 story's intent that creation is accessible from Settings at any time, not only from an empty state.

**Interaction** — Deliberate, correct divergence from `SettingsFocusSection`: the Focus section returns `null` when empty, while Routines always renders heading + divider + add row. That's intentional and right for a canonical management surface — don't reconcile it back to the Focus pattern.

**Minor observation (non-blocking, inherited pattern):** On unarchive, `useRoutines` data is not refetched, so the routine reads stale `archived: true` (excluded from active) while `locallyUnarchived` removes it from archived — it briefly appears in neither list until remount/refetch. This is the *identical* pattern already shipped in `SettingsFocusSection` (same stale-Set approach), so it's not introduced here and doesn't gate this PR. Worth a future follow-up to re-list the unarchived item in the active group rather than just hiding it.

Issue #122, opened 2026-04-08
Add Routines section to Tend dashboard


sfrankle:
## Story
As a user, I want to see my Routines on the Tend dashboard so that the ones I should do now are front and centre, and the rest are out of the way until I need them.

### Section structure
- [ ] Routines section appears above the Focus pills section and entry type tile grid
- [ ] Section is absent if the user has no active Routines (no empty state, no section header)
- [ ] Archived Routines do not appear

### Due-now cards
- [ ] Routines due in the current time block and not yet completed this block appear as full cards at the top
- [ ] Card shows: Routine name, time blocks, completion count for today (e.g. "1 of 3 · Morning done")
- [ ] Tapping a due card opens the complete-Routine flow

### Collapsed disclosure row
- [ ] A single collapsed row below active cards holds: completed Routines + not-yet-due Routines
- [ ] Disclosure row label reflects contents (e.g. "Completed · Later")
- [ ] Tapping expands to show these Routines — user can tap any to complete early
- [ ] Completed Routines show their completion count

## Notes
Completion tracking logic (what counts as completed in current block vs fully done for the day) is covered in a separate issue. This issue is about the dashboard layout and navigation into the complete-Routine flow.

This Routines section is added to a Tend dashboard that already has the Focus pills section from the Focus milestone.

PR #177, opened 2026-07-27
Refactor: extract getTypedDb() helper to eliminate repeated cast

sfrankle:
## Summary

- Every DB call site repeated `(await getDb()) as unknown as Db` just to get a typed handle. This adds a single `getTypedDb(): Promise<Db>` helper (`src/lib/db/typed-db.ts`) that owns the cast once.
- All 20+ call sites across log screens, tab screens, hooks, and components now call `getTypedDb()` instead of casting inline.
- **Pure refactor — no schema change, no migration, no runtime behaviour change.** Reviewers do not need to look for a migration test; nothing about the data layer's behaviour moved.

Closes #169

## What to review

- `src/lib/db/typed-db.ts` — the new helper. This is the one place the `as unknown as Db` cast now lives.
- The remaining diff is mechanical: swapping the inline cast for a `getTypedDb()` call at each site.
- `src/lib/db/__tests__/typedDb.test.ts` — two new tests covering the helper.

## Testing

- `npx tsc --noEmit` — clean
- `npx eslint . --ext .ts,.tsx` — 0 errors
- `npm test -- --ci` — 687 tests pass, including 2 new `typedDb` tests
- No behaviour change, so no Maestro flow updates needed.

## Checklist
- [x] Acceptance criteria met
- [x] Flow tests (Maestro) added or updated for any user-facing behavior
- [x] Migration test written if schema changed
- [x] CI passing
- [x] `docs/changelog.md` updated
- [x] No judgmental language in UI strings
- [x] No network calls or off-device data transmission introduced

Issue #160, opened 2026-05-22
Routine queries, hooks, and ScheduleableBlock utility


sfrankle:
## Summary
Data layer for the entire Routines feature: query functions, React hooks, and the `ScheduleableBlock` type used by scheduling and the time-block picker.

## Acceptance criteria
- [ ] `ScheduleableBlock = Exclude<TimeBlock, 'Night'>` type exported from `src/lib/utils/timestamp.ts`
- [ ] `getScheduleableBlocks(): ScheduleableBlock[]` utility returns `['Morning', 'Midday', 'Afternoon', 'Evening']`
- [ ] Query functions in `queries.ts`: `createRoutine`, `updateRoutine`, `setRoutineArchived`, `getRoutines` (active + archived flag), `getRoutineItems(routineId)`
- [ ] `getRoutineCompletionState(db, routineId, timeBlock, today)` → `'due' | 'completed_this_block' | 'fully_done'`; logic: count completions today via `routine_completion` join; compare against configured time block count and current block window
- [ ] `useRoutines(options?: { includeArchived })` hook
- [ ] `useRoutineCompletionStates(routines, currentTimeBlock, today)` hook (used by Tend dashboard)
- [ ] Unit tests for all query functions; completion state logic tested thoroughly — all examples from story #125 AC pass
- [ ] Tests: unit tests in `src/__tests__/db/queries-routines.test.ts`

## Notes
`getRoutineCompletionState` is the most complex piece in the milestone — test all four examples from #125 AC before marking done.

The `ScheduleableBlock` type constrains the time-block picker and the CHECK constraint in `routine_time_block` to Morning/Midday/Afternoon/Evening only.

Blocked by: #121

sfrankle:
## Implementation Plan

**Approach:** Add `ScheduleableBlock = Exclude<TimeBlock, 'Night'>` and `getScheduleableBlocks()` to `timestamp.ts`, then append a Routine section to `queries.ts` with six functions (`createRoutine`, `updateRoutine`, `setRoutineArchived`, `getRoutines`, `getRoutineItems`, `getRoutineCompletionState`) following the exact patterns from the Focus query layer. Two new hooks (`useRoutines`, `useRoutineCompletionStates`) follow the `useFocuses` pattern. No schema changes — v9 migration is already in place.

**What will change:**
- `src/lib/utils/timestamp.ts` — add `ScheduleableBlock` type and `getScheduleableBlocks()` utility
[...]

PR #178, opened 2026-07-27
Tend dashboard — Routines section

### LINKED ISSUE #163: Tend dashboard — Routines section

sfrankle:
## Summary

- Adds a Routines section to the top of the Tend dashboard, above the Focus pills row, so a Routine you're meant to do right now is the first thing you see when you open the app.
- Routines are split into three groups: **due now** (rendered as tappable cards), and **completed** / **later** (tucked behind a single collapsed disclosure row). A Routine whose block has already started stays available for the rest of the day — it is never "overdue" or "missed", and nothing in the UI implies you fell behind.
- The grouping and copy rules are pure functions in `src/lib/utils/routine-dashboard.ts` so the product semantics can be tested directly against the design spec rather than through the component tree. The component layer just renders what those functions return.

## What to review

**`src/lib/utils/routine-dashboard.ts`** — this is where the judgment calls live:

- `groupRoutinesForDashboard` — a Routine goes to `dueNow` once *any* of its configured blocks has started (not only the current one), which is what keeps a morning Routine reachable at 4pm without shaming language. A Routine with no configured blocks means "anytime" and is always due. A Routine whose completion state hasn't resolved yet (loading, or a failed read) is parked in `later`, never optimistically shown as due — I'd rather under-prompt than invent a task.
- `formatRoutineProgress` — the count is deliberately **not** clamped to the number of configured blocks, so a Routine completed 3 times against 2 blocks reads "3 of 2". Clamping to "2 of 2" would quietly hide something the user actually did. Returns `null` at zero completions; silence reads better than "0 of 2".
- `disclosureLabel` — the label reflects what's actually behind it ("Completed", "Later", or "Completed · Later"), and returns `null` so the row isn't rendered when empty.

**Also in this branch, worth flagging:**

- **Staleness fix, via a new shared hook shell `src/hooks/useRoutineKeyedRead.ts`** — navigating back to Tend after completing a Routine used to leave the section showing the pre-completion state: the routine IDs, time block and date are all unchanged, so the effect had no reason to re-run. Both dashboard hooks now sit on a shared shell that owns the dep-stability and focus-refresh contract, matching how `useRoutines`/`useFocuses`/`useTraceEntries` already work. (This landed first as an opt-in `refreshKey` parameter; review made the case that an *optional* refresh param is an opt-in staleness bug — a future caller that omits it type-checks cleanly and silently reproduces the original defect — so the refresh moved inside the hook and the parameter is gone.)
- **Maestro flows un-stubbed** — `routine-complete.yaml` and `routine-create-edit.yaml` were placeholders and are now real flows; `routine-dashboard.yaml` is new. They share a new `maestro/subflows/create-morning-routine.yaml` so each flow can set up its own Routine without depending on ordering.

## Testing

- Unit tests for the grouping and copy helpers (`routine-dashboard.test.ts`) covering every group transition, the "anytime" case, the unresolved-state case, the un-clamped progress count, and all disclosure label permutations.
- Component tests for `RoutineSection` covering the empty state (`+ Add Routine` only), due-now cards, archived Routines being excluded, and disclosure row expand/collapse including completing a `later` Routine early.
- Hook tests for `useRoutineDayProgress` and `useRoutineCompletionStates`, including a focus-driven re-read guarding the staleness trap.
- `RoutineSection` tests for the `Night → Evening` state mapping, and for a Morning Routine still reading as due (never missed) at night.
- A query test pinning the invariant that keeps an unclamped "3 of 2" off a due-now card: `fully_done` fires once completions ≥ configured blocks, so an over-count can only appear inside the collapsed disclosure.
- Query tests for the batched `getRoutineDayProgress` read.
- `TendScreen.test.tsx` updated for the new section.
- Maestro: `routine-dashboard.yaml` (new), plus `routine-complete.yaml` and `routine-create-edit.yaml` now exercising real behaviour.

Note: the repo's Jest suite is currently intermittently flaky on `main` (unrelated transaction-rollback tests under jest-expo's worker setup), so a red CI run may not reflect this branch.

## Checklist
- [x] Acceptance criteria met
- [x] CI passing

Closes #163
Contributes to #122


sfrankle (commented):
## Simplify Pass

Four cleanup agents (reuse, simplification, efficiency, altitude) reviewed the branch diff. Three of the four independently converged on the same top finding, so that is where most of the change is.

### Fixed

**The two dashboard hooks were one hook written twice.** `useRoutineDayProgress` duplicated ~60 lines of `useRoutineCompletionStates` — the `routinesRef` mirror, the id-join dep string, the `cancelled` flag, the loading/error triad, and a lightly-reworded copy of the same explanatory comment. Extracted the shared shell to `src/hooks/useRoutineKeyedRead.ts`; both hooks are now ~40 lines of just their own query.

**`refreshKey` was an opt-in staleness bug.** The altitude pass made the case that a caller-supplied refresh counter is the wrong depth: every sibling data hook (`useRoutines`, `useFocuses`, `useTraceEntries`) owns its focus reload internally, and an *optional* parameter means a future consumer that omits it type-checks cleanly, runs cleanly, and silently reproduces the exact bug the parameter was added to fix. The focus tick now lives inside `useRoutineKeyedRead`; `refreshKey` is gone from both public signatures and `RoutineSection` no longer plumbs it. The first focus is skipped deliberately — it is the mount, which the read effect already covers, and bumping on it would fire a duplicate query on every mount.

**The collapsed disclosure row was a hand-rolled second `RoutineCard`** — same `accessibilityRole`, same `Complete ${name}` label template, same progress guard. It is now `<RoutineCard variant="compact">`, so the press target and a11y label have one definition.

**`BLOCK_ORDER` restated the block order** ten lines from an `inBlockOrder` that derived it from `getScheduleableBlocks()`. Now derived from the same source — but resolved lazily and cached, not at module load. My first attempt computed it at import time and broke 8 unrelated suites that partially mock `timestamp.ts`; the lazy form keeps the single source of truth without running against a mock in files that never touch a Routine.

**Smaller:** `getRoutineDayProgress` aggregated through three parallel structures with two unreachable `?? fallback` branches — collapsed to one seeded tally. The `labelMedium` font triple was spelled out five times across the two components — now one composed base. `dayjs().format('YYYY-MM-DD')` inlined in a component became `todayLocalDate()` in `timestamp.ts`. `EMPTY_PROGRESS` guarded a case the query rules out — `formatRoutineProgress` takes `undefined` instead. `routine-create-edit.yaml` still inlined the create steps the shared subflow was extracted to replace; it now calls the subflow, and its second item moved to the edit screen, which covers add-item-on-edit rather than repeating the create path.

### Skipped

**The two hooks still issue two reads of `routine_completion` for the same routines on the same day.** Three agents flagged this and they are right: `getRoutineCompletionState` fans out 2–3 queries per routine per focus, while `getRoutineDayProgress` already answers a superset in one batched query. Everything the first needs is derivable from the second plus `routine.timeBlocks`. Collapsing them would remove the N+1, make state derivation unit-testable without a DB, and close the window where a card's progress line and its group placement come from two different reads. Not done here because it reshapes `getRoutineCompletionState`, a tested API from #160 — that is a ticket, not a cleanup pass.

**The `Night → Evening` substitution stays in the component.** Night handling now has three answers across three layers, and the UI one is the only one that produces a wrong result (a zero-block Routine completed 22:00–04:59 reads as due again). The real fix is teaching `timeBlockWindow` about Night, which is out of scope for a cleanup pass and is a correctness change, not a simplification.

**The defensive `!r.archived` filter stays** — it is deliberate, requested by the plan critic, and has a test asserting an archived fixture never renders. It is now memoised rather than reallocating each render.

### Verification

`npx tsc --noEmit` clean. `npx eslint . --ext .ts,.tsx` — 0 errors, 10 warnings, all pre-existing. `npm test -- --ci` — 743 passed, 743 total on a clean run, same count as before the pass.

The suite is intermittently flaky and it is not this branch: across repeated full runs I saw 3, 1, and 0 failures, always the same transaction-rollback family (`queries-focus`, `queries-routines`), each passing in isolation. The implementer measured the same pattern on `main` with this work stashed. Worth its own issue.

sfrankle (commented):
## Safety Review

**Verdict:** PASS

### Privacy: OK
No `fetch`/`axios`/HTTP client, no WebSocket, no analytics/crash-reporting/telemetry imports anywhere in the diff. No `package.json` change — no new dependencies, no new `expo-*` package, no new permissions. The only URL introduced in the whole diff is the GitHub PR link in the `docs/changelog.md` row.

**Privacy language:** clean. Grepped the branch diff, the working-tree diff, all 11 commit messages, the PR body, `docs/changelog.md`, and `docs/decisions.md` for clinical/community terminology — zero matches. Nothing prohibited reaches GitHub or any checked-in file.

### Data Safety: OK
No schema migration in this PR, and none is required — confirmed by inspection rather than by absence:

- No file under a migrations path is touched.
- The only production SQL added is `getRoutineDayProgress` in `src/lib/db/queries.ts`, a single `SELECT routine_id, created_at FROM routine_completion WHERE routine_id IN (...) AND substr(created_at, 1, 10) = ?`. Read-only.
- Every `INSERT`/`DELETE` in the diff is confined to `src/__tests__/db/queries-routines.test.ts` fixture setup. No production write path, and no seed-data insert, so the `INSERT OR IGNORE` rule does not apply here.
- No column drop, table drop, or destructive statement anywhere.

The `IN (...)` list is built from `?` placeholders with the ids bound as parameters, not string-interpolated — no injection surface. Every requested id is seeded into the result map, so callers never hit an undefined key.

### Judgmental Language: OK
Every user-facing string was checked at the source rather than by grep alone. The rendered copy is: `Anytime`, `Completed`, `Later`, `Completed · Later`, `Once today` / `N times today`, `N of M · Morning, Afternoon done`, `+ Add Routine`, and the accessibility label `Complete {name}`. Nothing penalising.

Grep hits for `overdue` / `missed` / `behind` / `streak` / `failed` all resolve to non-user-facing text:
- Doc comments and the `docs/decisions.md` row explaining the rule *against* that vocabulary
- Test names and the guard assertion `expect(output).not.toMatch(/overdue|missed|skipped|behind|streak/i)`
- A `db failed` string inside a test error fixture
- `behind` used spatially ("behind the disclosure row"), not evaluatively

Worth noting for the record: the guard test in `src/lib/utils/__tests__/routine-dashboard.test.ts` makes this rule enforced rather than merely documented, and `formatRoutineProgress` returning `null` at zero completions avoids a "0 of 2" empty state. The un-clamped "3 of 2" is the right call — clamping would hide work the user actually did.

### Workflow Artifacts: OK
- `Closes #163` — present
- `Contributes to #122` — present
- Milestone `Routines` (#9) — assigned
- `docs/changelog.md` — exactly 1 new row, uses PR number 178 (not the issue number) as a markdown link to the full PR URL, inserted at the top per the file's stated convention
- `docs/decisions.md` — 1 new row appended to the **end** of the table, after the #177 row; nothing inserted mid-table
- `CI passing` checkbox — correctly left **unchecked**. `Migration test written if schema changed` also unchecked, which is correct given there is no schema change.

### Working tree (included per request)
The two staged-but-uncommitted Maestro files (`maestro/flows/tend/routine-create-edit.yaml`, `maestro/subflows/create-morning-routine.yaml`) are a test-only dedup: the inline create steps are replaced by a `runFlow` call to `maestro/subflows/create-morning-routine.yaml`, a caller comment is added to the subflow, and an add-item step is added on the edit screen. No privacy, data-safety, or language impact. They do still need to be committed before merge.

All safety checks passed.

sfrankle (commented):
## Code Quality Review

**Verdict:** CONCERNS

Type-check, lint, and the full unit suite for the touched files all pass (83 tests). The pure/hook/component layering is clean and the `/simplify` pass landed well — `useRoutineKeyedRead` removes the opt-in `refreshKey` footgun, and lazily caching `BLOCK_ORDER` behind a function is the right call given the partial `timestamp.ts` mocks in other suites. The main issue is in the newly-activated Maestro flows.

### Maestro / E2E

**1. `tapOn: id: "tend-tab"` targets a testID that does not exist — blocking for the flows themselves.**
`grep -rn "tend-tab" app src` returns nothing; `app/(tabs)/_layout.tsx` sets no `tabBarTestID` on any `Tabs.Screen`. This was harmless while the flows were commented-out stubs, but all three are now live and step 1 (or step 4, in `routine-create-edit.yaml`) will fail on every run:

- `maestro/flows/tend/routine-dashboard.yaml:15`
- `maestro/flows/tend/routine-complete.yaml:15`
- `maestro/flows/tend/routine-create-edit.yaml:15,55`

The repo convention is a text tap on the tab label — `maestro/flows/settings/settings-routines.yaml` uses `- tapOn: "Settings"`, and this PR already uses that form for Settings in `routine-create-edit.yaml`. Either switch to `- tapOn: "Tend"` or add `tabBarTestID: "tend-tab"` in `app/(tabs)/_layout.tsx`. Since Maestro is not in CI, nothing else will catch this.

**2. Stale comment referencing the removed refresh key.**
`maestro/flows/tend/routine-dashboard.yaml` — "a stale due card here would mean the refresh key regressed". `refreshKey` was deleted by the simplify pass; the mechanism is now the focus tick inside `useRoutineKeyedRead`.

**3. Subflow comment is inaccurate about the pre-05:00 case.**
`maestro/subflows/create-morning-routine.yaml` says a Morning Routine is "outside that window still reachable via the Later disclosure row". Before 05:00 the current block is `Night`, and `groupRoutinesForDashboard` treats every configured block as started at Night — so the card is due-now at literally every hour. The flow assertions are correct; the comment describes a branch that cannot be taken.

### Tests

**4. Night → Evening substitution is untested (flagged by the plan critic).**
`RoutineSection.tsx` maps `nowBlock === "Night"` to `Evening` before handing the block to `useRoutineCompletionStates`, and the surrounding comment documents a real accepted edge case. `RoutineSection.test.tsx` pins `getTimeBlock` to `Morning` for every test, so the substitution has no coverage; `routine-dashboard.test.ts` covers Night for *grouping* only. One test that re-mocks `getTimeBlock` to `Night` and asserts `useRoutineCompletionStates` was called with `Evening` (and that grouping still puts the Routine in due-now) would lock in both halves.

Otherwise the test quality is strong: names describe behaviour, empty/loading/read-failure states are all covered, the voice guard test is a nice touch, and the `useFocusEffect` mocks are faithful (once per focus, not once per render) rather than papering over the real hook.

### React Native / Expo

**5. `useRoutineKeyedRead` writes refs during render.**
`routinesRef.current = routines` and `loadRef.current = load` execute in the render body. It works today (the effect only reads them after commit) but it is a render side effect, which React documents as unsafe under concurrent rendering. Assigning inside a `useEffect` with no dep array, or a small `useLatest` helper, is the conventional form.

**6. Stale data survives a dep change.**
When `routineIds` or `depKey` changes, `data` is left holding the previous result while `loading` flips to `true`. `RoutineSection` ignores `loading` from these hooks and reads `progress[routine.id]` directly, so immediately after a date rollover on focus a card can briefly render yesterday's progress line. Clearing `setData({})` at the top of the effect (or gating the progress text on `loading`) closes it.

**7. First-focus skip assumes mount === focus.**
`isInitialFocus` swallows the first `useFocusEffect` call on the premise that mount already triggered the read. If the component ever mounts while unfocused (tab preloading, `lazy: false`), the first genuine focus is swallowed and the section stays stale until a second focus. Expo Router tabs are lazy by default so this is latent, not live — worth a comment at minimum.

### TypeScript

No findings. No `any`, no assertions hiding real errors, props interfaces defined on both new components, exported functions have explicit return types, and `RoutineDayProgress` is a well-documented public shape.

### Haven Patterns

No findings. `docs/decisions.md` has a new row appended at the end covering the due-now grouping rule, the no-overdue vocabulary, and the deliberate unclamped "3 of 2" count. Changelog row present and skimmable. `getRoutineDayProgress` follows the existing `substr(created_at, 1, 10)` date-comparison and parameterised-`IN` conventions from `getRoutineCompletionState`, seeds every requested id so callers never branch on a missing key, and lives in the query layer with no direct `db.*` calls from screens. The disclosure chevrons reuse the existing `▲`/`▼` glyphs and the `fontSize: 12` / `colors.chrome` styling from `app/(tabs)/(tend)/routine/[id]/complete.tsx` — no new glyph pair introduced, as requested.

### General

No dead code, no `console.log`, no commented-out blocks. Note that the two Maestro files with staged-but-uncommitted dedup edits still need committing before merge.

---

None of this is a correctness problem in shipped app code — items 1–4 are test/flow gaps, 5–7 are latent robustness. Item 1 should be fixed before merge since the flows are advertised as un-stubbed.

sfrankle (commented):
## Product Vision Review

**Verdict:** CONCERNS

Read against user story #122, technical task #163, `docs/superpowers/specs/2026-04-08-focus-routines-notes-design.md`, `docs/design/interaction.md`, and `docs/design/brand.md`. The core principle — a Routine is never overdue, missed, or behind — is honoured thoroughly and defensibly. The concerns below are interaction and accessibility polish, not vision failures.

### User Story Fulfillment

Every acceptance criterion in #122 is met in spirit, not just technically.

- Section sits above the Focus row in `ListHeader`; archived Routines are filtered twice (hook default plus a defensive filter in `RoutineSection`).
- The broadening of "due now" from *current block* to *any block that has started* is the right reading of the story. The literal criterion ("due in the current time block") would have made a Morning Routine disappear at 13:00 — which is exactly the shaming behaviour the story's final criterion forbids. Choosing the spirit over the letter here is correct, and it matches the spec's worked example at line 157.
- Unresolved completion state parks a Routine in `later` rather than optimistically presenting it as due. Under-prompting over inventing a task is the right instinct for Haven.
- The guard test asserting no output string matches `/overdue|missed|skipped|behind|streak/i` is a good durable protection for a principle that would otherwise erode one copy tweak at a time.

Minor divergence worth noting, not fixing: the design spec says "If a user has no Routines, this section is absent," while #122/#163 specify a `+ Add Routine` row. The implementation follows the issues. That's the better call — it mirrors `+ Add Focus` and keeps the feature discoverable — but the spec now disagrees with shipped behaviour.

### UX Principles

**Empty state — passes.** A single `+ Add Routine` affordance, no illustration, no explanatory apology, no "You haven't set up any Routines yet." It reads as an available door rather than an unfilled slot, and it's identical in shape to the Focus row's empty state, so a new user sees one consistent pattern rather than two different flavours of nothing. Rendering it even while reads are in flight is a nice touch — the create path never depends on a query succeeding.

**Error state — passes, with a note.** No error UI at this surface; a failed read degrades to `+ Add Routine`. That's appropriately quiet. The note: on a read failure, `useRoutineKeyedRead` sets `data` to `{}`, so every Routine falls into `later` and appears under a **"Later"** disclosure. "Later" is a positive assertion about scheduling that may be untrue — a Routine that is in fact due now is being described as not-yet-due. It's not alarming and the Routine is still one tap away, so this isn't blocking, but a Routine silently relabelling itself is a small honesty cost. Grouping unresolved Routines into an unlabelled group (or leaving the label off when nothing is confidently classified) would avoid asserting anything.

**"3 of 2" — the surrounding design does keep this from reading as a score.** Not re-litigating the decision, only assessing the framing, and the framing holds up:

- The over-count is structurally unreachable on a due-now card. `fully_done` triggers once completions ≥ configured blocks, so any Routine where N ≥ M is already in `completed` and behind the collapsed row. "3 of 2" is therefore only ever visible *after* the user has deliberately expanded a disclosure — it is never on the prominent card face, and never on the first thing seen when opening the app. That's a meaningful mitigation and it's worth stating explicitly somewhere, because it's currently an emergent property of the grouping rules rather than an intentional invariant anyone would know to preserve.
- The progress line is plain `chrome`-coloured metadata at label size, adjacent to the block list, with no bar, ring, tick, or percentage. Nothing reads as a gauge.
- Returning `null` at zero completions is the single most important copy decision in this file. "0 of 2" on a fresh morning card would have been a deficit staring at the user; silence is right.

**Progress copy — passes.** "Once today" / "3 times today" are warm and observational rather than evaluative — they describe what happened, not how it measures up. "Anytime" is a good non-answer to "when is this due" and avoids inventing a deadline for an unscheduled Routine. "1 of 2 · Morning done" reads as a record, and "done" is the right word — it's factual and terminal without being congratulatory.

**Disclosure row — reads calm.** "Completed", "Later", "Completed · Later" are neutral nouns with no count badge, no urgency colour, no chevron pointing at unfinished work. Returning `null` so the row disappears entirely when empty matters — a permanently-present "Completed · Later" row with nothing behind it would read as a container waiting to be filled. Collapsed by default and mixing done-things with not-yet-things in one bucket is what stops it becoming a backlog: a to-do list separates what's left, and this deliberately doesn't.

One thing does undercut that, below.

### Interaction

**1. Expanded disclosure gives no signal for which Routines are done.** Inside the expanded list, completed and later Routines render as identical `compact` rows — name plus optional progress text. The only distinction is whether a progress line happens to be present. It works out (completed Routines always have count > 0, later Routines effectively never do), but it's an implicit signal carried by the *absence* of a line, which is the weakest kind. A user scanning the expanded row can't tell at a glance what they've already done, and may re-tap a completed Routine — which is the most likely real-world route to generating a "3 of 2". Worth considering a lightweight separation (a quiet "Completed" / "Later" subheading when both groups are present) so the disclosure stays a record rather than an ambiguous list.

**2. Tap target on `+ Add Routine` (`RoutineSection.tsx`).** The `Pressable` wraps only the `Text`, inside a padded `Surface`. The row looks like a card but only the text glyphs respond — the 12dp of surface padding around it is dead. Compare `RoutineCard`, where `Pressable` wraps `Surface` and the whole card is live. The resulting target is roughly text height (~20dp), against the 48dp Material minimum called out in `docs/design/brand.md`. This is the *primary* affordance in the empty state, so it's the one that most needs to be forgiving. Moving `Pressable` outside `Surface` (matching `RoutineCard`) fixes both the dead zone and the size.

The disclosure row is borderline too — 12dp vertical padding either side of a label line lands around 44dp.

**3. Screen reader loses the card content.** `RoutineCard` sets `accessibilityLabel={`Complete ${routine.name}`}` on the `Pressable`, which collapses the subtree — the time blocks line and the progress line are not announced. A VoiceOver user gets "Complete Morning Flow, button" and no sense of when it's scheduled or what they've already done today, which is precisely the information this card exists to convey. Folding the block and progress text into the label would restore it. Related: completed Routines in the disclosure also announce "Complete X", which is confusing when the thing is already done — though the action genuinely is "complete again", so some phrasing thought is needed rather than a mechanical fix.

**4. First-load reshuffle.** `RoutineSection` gates rendering on `useRoutines().loading` only, not on completion-state loading. In the window after Routines resolve but before states do, every Routine is unresolved and lands in `later` — so a "Later" disclosure row renders, then vanishes as cards pop in above it. `docs/design/interaction.md` asks for "stable layouts without abrupt movement", and this is the very first thing the user sees on opening the app. Focus refreshes are unaffected (prior `data` is retained), so this is cold-start only, but cold start is the moment that sets the tone. Holding the section until states resolve would avoid showing a grouping the code doesn't yet believe.

### Tone

Clean. No scores, no streaks, no ranking, no time pressure, no motivational voice. Nothing counts down, nothing turns a colour, nothing congratulates. "Completed", "Later", "Anytime", "done", "Once today" — all observational. The changelog entry ("what's due now up top, completed and later ones tucked behind a tap") carries the same voice.

### Side Effects

- `useRoutineCompletionStates` was rewritten onto `useRoutineKeyedRead`. `RoutineSection` is its only non-test consumer, so the blast radius is contained and the public shape is unchanged.
- The focus-refresh fix is a genuine user-visible improvement beyond the ticket: without it, completing a Routine and returning to Tend left the card sitting there as though nothing had happened — which would have read as data vanishing. Good catch to fold in.
- The Night → Evening fallback edge case (an unscheduled Routine completed between 22:00 and 04:59 reading as due again) is documented in-code and accepted. From the user's side this surfaces as a Routine reappearing after they did it. Low frequency, non-destructive, and honestly flagged — fine to carry, but it belongs in a follow-up issue rather than only in a code comment.
- No unrelated flows touched. `+ Add Focus`, the tile grid, and the complete-Routine screen are unchanged, and "Saved." on the complete screen means the card disappearing from the dashboard on return is already acknowledged.

---

None of the above changes the shape of the feature. The vision-critical work — no penalty language, no deficit framing, an empty state that isn't a failure, and a disclosure that doesn't become a backlog — is done and done well. Items 2 and 3 under Interaction are the ones I'd want addressed before merge; 1 and 4 are judgement calls the owner may reasonably defer.

sfrankle (commented):
## Code Quality Review (re-run)

**Verdict:** PASS

All three concerns from the first pass are resolved, and each fix is the right shape rather than the minimum that silences the finding.

### 1. `tend-tab` testID — resolved
Verified: `tend-tab` appears nowhere in `app/` or `src/`, so all four taps were dead on arrival. `- tapOn: "Tend"` now matches the pre-existing convention in `maestro/flows/navigation/tab-navigation.yaml` and `log-screen-back-navigation.yaml`, not just `tapOn: "Settings"` — good, that makes it the established pattern rather than a new one. The `routine-create-edit.yaml` dedup onto `subflows/create-morning-routine.yaml` landed too, and replacing the duplicated create-screen item steps with the *edit* screen's add-item path is a net coverage gain, not just a line count reduction.

### 2. Night → Evening — resolved
Two tests, and they assert the two different things that can break. The first pins the `stateBlock` mapping at the boundary the type system can't express (`ScheduleableBlock` excludes Night); the second asserts the product rule — a Morning Routine at 22:00 is still a due-now card, never behind the disclosure. Both are behaviour assertions with names that say what the code does. `mockGetTimeBlock.mockReturnValue('Morning')` in `beforeEach` removes the ordering coupling the suite had before.

### 3. `isInitialFocus` skip — resolved
Removing it outright is the correct call, and the replacement comment in `useRoutineKeyedRead.ts` records the reasoning (tab pre-render decouples mount from focus; a duplicate local SQLite read is the cheaper error) so it doesn't get reintroduced as an optimisation later.

The test rewrite deserves a specific note: swapping `toHaveBeenCalledTimes(1)` for a loop asserting every call carries `[1, 2]` tests the property that actually matters — batching — instead of pinning a call count that is an artifact of the refresh strategy. Same for the focus test using `toBeGreaterThan(callsBefore)`. That is the behaviour-over-implementation line in the right place.

### Deliberate non-changes — I agree with both

**Refs assigned during render.** Confirmed pre-existing: `useRoutineCompletionStates.ts` at the base SHA did `routinesRef.current = routines` at render scope. The extraction preserved it rather than introducing it, and the latest-value semantics are load-bearing for the effect body. Changing it belongs in its own change, if ever.

**`data` not cleared on `depKey`/`routineIds` change.** Agreed, and the reasoning is the stronger one: a blank-then-repopulate is a visible flash on a screen the user is actively looking at, versus a sub-second stale progress line after a date rollover. Correct tradeoff for this surface.

### Also noted in this push
- `RoutineCard` accessibility label now composes name + blocks + progress. This is a real fix I had not flagged — `Pressable` collapses its subtree for screen readers, so the block and progress `Text` lines were previously unreachable. The comment explains why the label has to carry them.
- Add-row `Pressable` moved outside `Surface`, so the tap target is the card rather than the text inside its padding. Now consistent with `RoutineCard`.
- `disclosureIcon` `fontSize: 12` is hardcoded, but it is byte-identical to the existing `expandIcon` style in `app/(tabs)/(tend)/routine/[id]/complete.tsx`. Matching the established pattern is right; tokenising the caret size is a separate cleanup across both call sites, not this PR's job.

### Verification
- `npx tsc --noEmit` clean
- `npx eslint` clean on all six changed source files
- 66 tests passing across the five affected suites
- No `console.*`, `TODO`, `any`, `as unknown`, or commented-out blocks in the diff
- `docs/decisions.md` entry appended at the end of the table, with the due-now grouping rule and the copy guard test recorded

Nothing outstanding. Good to merge once CI is green.

sfrankle (commented):
## Product Vision Review (re-review — 5ecbd85)

**Verdict:** PASS

Both concerns from the first pass are addressed. Two optional notes below, one of which is a one-line change I would take before merge.

### Previously raised — now resolved

**`+ Add Routine` tap target.** `RoutineSection.tsx` now wraps the `Surface` in the `Pressable` rather than the `Text`. The add row is now the same size of target as a `RoutineCard`, which is what a user reaches for. Resolved.

**`RoutineCard` accessibility label.** The composed label (`Complete {name}. {blocks}. {progress}`, with the blocks line dropped in the compact variant) means a screen reader user now hears the same three facts a sighted user reads, in the same order, instead of just the name. The comment above it explains *why* the label has to carry the subtree, which is the part a future editor would otherwise break. Resolved.

### User Story Fulfillment

All of #122's criteria are met in spirit, not just literally. Section sits above `FocusRow` in `app/(tabs)/(tend)/index.tsx`; archived Routines are filtered defensively even though `useRoutines` already excludes them; due-now cards carry name, blocks and progress; one disclosure row holds completed + later with a label that names its actual contents; anything inside it can still be completed early.

The grouping rule is the thing that makes this feel like Haven rather than a task list: a Routine becomes due once *any* configured block has started and stays there for the rest of the day. A Morning Routine at 20:00 is still simply available. That is the right call and it is now written down in `docs/decisions.md`.

### UX Principles / Tone

No scores, no bars, no ticks, no percentages — completion is plain text. The voice guard test in `routine-dashboard.test.ts` pins the absence of overdue/missed/skipped/behind/streak across every formatter output, which is a better protection than a review comment.

Empty state is right: no Routines yields the add row alone, with no "you have no Routines yet" framing. Error state is right too — a failed read degrades silently to the add row rather than showing an alarm, and the create path never depends on a read succeeding.

`formatRoutineProgress` returning `null` at zero completions is the correct instinct. "0 of 2" would read as a deficit; silence reads as nothing has happened yet.

### On the over-count ("3 of 2")

Still structurally unreachable on a due-now card, confirmed against 5ecbd85. `getRoutineCompletionState` returns `fully_done` whenever `blockCount > 0 && completionCount >= blockCount`, so a card can only reach `dueNow` with `completionCount < blockCount` (or `blockCount === 0`, which routes to the "Once today" / "N times today" branch instead). The over-count is only ever displayed inside the disclosure, on a Routine the user completed extra times on purpose — where showing the true count is the honest thing to do.

Worth pinning, yes, and cheaply. `queries-routines.test.ts:479` already covers the `2 >= 2` boundary. I would add one sibling case — three completions against two configured blocks still returns `fully_done` — and one line on `formatRoutineProgress`'s doc comment noting that the unclamped count is reachable only from the disclosure, upheld by that test. That way if someone later relaxes the `>=` to `>`, or adds a `dueNow` path that bypasses the state check, a test fails rather than a card quietly reading "3 of 2".

### Interaction — the two deliberate non-changes

**Completed vs later in the expanded disclosure — agreed, leave it.** This is a design decision, not a review fix, and inventing a divider or per-row status treatment mid-review would be worse than leaving it. Recording it for Sarah: under a bare "Completed · Later" label, a `completed_this_block` Routine reads as e.g. "Morning Flow — 1 of 2 · Morning done", which a user could reasonably parse as belonging to the Later half. If it ever gets treatment, the cheapest version that invents no new vocabulary is two small chrome-coloured subheadings reusing the existing "Completed" / "Later" words and `disclosureLabel` styling, shown only when both groups are non-empty. Not blocking — the disclosure is secondary content and every row in it is tappable either way.

**Cold-start reshuffle — this one I would change, and it costs nothing.** On first render `states` is empty, so every Routine falls into `later`, so the disclosure renders as "Later" and then reshuffles as cards pop in above it. The cards cannot appear any sooner regardless, since `dueNow` is empty until the states read resolves — so gating the *disclosure only* on completion-state loading adds zero latency to the cards and removes a transient label that is actively wrong:

```ts
const { states, loading: statesLoading } = useRoutineCompletionStates(routines, stateBlock, today);
// ...
{!loading && !statesLoading && label !== null && ( ... )}
```

The section then goes: add row → cards + disclosure, instead of add row → "Later" → cards + corrected disclosure. Small, but "Later" flashing before a due-now card is the kind of flicker that makes a calm surface feel unsettled.

### Side Effects

The `refreshKey` / `useRoutineKeyedRead` change is in-scope repair, not scope creep — without it a user completing a Routine and navigating back to Tend would see it still sitting in the due list, which is exactly the "data behaves unexpectedly" failure this review looks for. The un-stubbed Maestro flows touch no product surface. Nothing outside the Routines section changed.

sfrankle (commented):
## Product Vision Review — focused re-review (group headings, 6d6a546)

**Verdict:** PASS

Scope: only the `Completed`/`Later` subheadings added inside the expanded disclosure. Everything else was reviewed previously and is unchanged.

### User Story Fulfillment
The specific problem raised earlier is solved. With both groups present, a completed Routine showing `1 of 2 · Morning done` now sits under an explicit `Completed` heading rather than under a bare `Completed · Later` row where progress text was the only (ambiguous) signal. A Later Routine carrying progress text is equally unambiguous now. No regression to the collapsed state — headings render only when expanded, covered by a test.

### UX Principles
Does not make the disclosure feel like a structured to-do list. Two chrome-coloured lines inside an already-collapsed region is close to the minimum intervention; there are no counts, no ticks, no ordering language, and the rows remain tappable-to-complete-early, so `Later` still reads as "ahead of you" rather than "outstanding". `Completed` before `Later` matches the disclosure label order — right call for consistency.

**The "only when both groups" rule is correct.** With one group the disclosure label already reads `Completed` or `Later`; a lone subheading two lines below would be pure restatement, and adding a second piece of chrome for zero information is exactly the kind of structure that starts making the section feel like a list. Keep it as is — do not make headings unconditional.

### Tone
`Completed` / `Later` reuse the existing disclosure vocabulary verbatim. No new words introduced, none of the forbidden vocabulary (missed/overdue/behind/incomplete). Good.

### Interaction — one non-blocking nit
`groupHeading` has `marginBottom: spacing.elementGap` but no `marginTop`, and `RoutineCard` compact rows carry `marginBottom: spacing.elementGap`. So the `Later` heading sits equidistant (12dp above, 12dp below) between the last Completed row and the first Later row — it does not visually attach to the group it labels, which slightly undercuts the grouping it was added to create. Suggest `marginTop: spacing.elementGap` on the heading (total 24dp above, matching the canonical major-group gap in `interaction.md`); the first heading is already fine because the disclosure row supplies its own `paddingVertical`.

Secondary: `Completed` appears twice within two lines when expanded (`Completed · Later` then `Completed`), in identical `labelMedium`/chrome typography. Acceptable — the repetition is what makes the grouping legible — but the extra top gap above would also help the eye read the second one as a subheading rather than an echo.

### Accessibility
Plain `Text` is consistent with the rest of the codebase (no `accessibilityRole="header"` is used anywhere today), and linear reading order means a screen reader user does hear `Completed` before the rows it covers, so the disambiguation survives. Worth noting that `RoutineCard`'s `accessibilityLabel` does not include the group, so the heading is the only cue — adding `accessibilityRole="header"` is a one-word change that enables rotor navigation and costs nothing. Recommended, not required.

### Side Effects
None. Collapsed rendering, disclosure label logic, and the loading gate are untouched; the change is a pure render-order refactor plus two conditional `Text` nodes. `showGroupHeadings && completed.length > 0` is redundant (`showGroupHeadings` already implies both are non-zero) — cosmetic only.

Implementation aligns with Haven's product vision. Recommend picking up the heading `marginTop` before merge; the rest is optional.

PR #181, opened 2026-07-28
fix: normalise better-sqlite3 errors in the test adapter to stop rollback-test flake

### LINKED ISSUE #179: Jest suite is intermittently flaky in the transaction-rollback tests
## Summary

The Jest suite fails intermittently on `main`, in a consistent family of tests, at a rate high enough that a red CI run is currently ambiguous.

## Evidence

Measured across repeated full runs on `main` (and independently on the #178 branch, with and without that work stashed):

| Run | Failures |
|---|---|
| 1 | 0 |
| 2 | 3 |
| 3 | 0 |
| 4 | 4 |

The failures are always in the same family — transaction-rollback assertions such as `routine query layer › createRoutine › rolls back routine row when items fail` and `query layer › saveEntry › transaction rollback on failure`, in `src/__tests__/db/queries-focus.test.ts` and `src/__tests__/db/queries-routines.test.ts`.

**Every one of them passes when the suite is run in isolation**, e.g. `npm test -- --ci --testPathPattern "queries-routines"`.

## Suspected cause

`better-sqlite3` under `jest-expo`'s multi-project worker setup. Not shared database files — `openTestDb` uses `:memory:`.

## Why this matters

Data-layer rollback behaviour is exactly the kind of thing we cannot afford to have failing noisily-but-ignorably. Right now the honest reading of a red CI run is "probably the flake", which is how a real rollback regression gets waved through. This undermines every PR, not just Routines work.

## Acceptance criteria

- [ ] Root cause identified and documented
- [ ] Full suite passes reliably across 10 consecutive runs
- [ ] No test weakened or skipped to achieve it — the rollback assertions must still fail when rollback genuinely breaks (verify by deliberately breaking one)

## Notes

Not milestone-scoped — this is test infrastructure and blocks nothing in particular, but degrades confidence everywhere. Surfaced during PR #178.

sfrankle:
## Summary

- The Jest suite failed intermittently in the sqlite transaction-rollback tests with "Received function did not throw". Rollback was never actually broken — the error was being misreported, which made a real rollback regression and this flake look identical in CI output.
- Root cause is a realm mismatch. better-sqlite3's `SqliteError` calls `Error.call(this)` instead of `super()`, so instances lack the `[[ErrorData]]` slot and `Object.prototype.toString.call(err)` returns `'[object Object]'`. The native addon binds one error constructor process-wide (`better-sqlite3/lib/database.js:57`), but Jest gives each test file its own vm context. Whichever db test file constructs a `Database` first binds its realm's `SqliteError`; later files then receive errors whose prototype chain reaches a foreign `Error.prototype`, so `err instanceof Error` is false there. Jest's `toThrowMatchers.js:47` gates on `isError`, so `rejects.toThrow()` reported no throw even though the promise had rejected.
- Fix normalises errors at the test-double boundary in `src/__tests__/db/adapter.ts` via a `toRealmError` helper, applied to all four adapter methods. No production code, schema, migration, jest config, or CI changes.

## What to review

- **`src/__tests__/db/adapter.ts`** — the `toRealmError` helper and its application across `getAllAsync`, `getFirstAsync`, `runAsync`, and `withTransactionAsync`. `message`, `name`, and `code` are preserved.
- **Why this isn't weakening a test.** The double emulates expo-sqlite, which rejects with genuine `Error` instances. Normalising here closes a fidelity gap between the double and the thing it stands in for, rather than papering over a failure.
- **Why the guard checks shape, not `instanceof`.** It tests `Object.prototype.toString.call(err) === '[object Error]'`. `instanceof` is precisely the value that varies with test-file load order, so it can't be the thing the guard depends on.
- **`src/__tests__/db/adapter-errors.test.ts`** — new regression guard asserting adapter rejections are genuine `Error`s.
- One row appended to `docs/decisions.md`.

## Testing

Rollback itself was verified to have never been broken: a 2000-iteration probe produced 2000 rejections and 0 leftover rows.

- **Deterministic pre-fix repro.** `npx jest --ci --runInBand --no-cache --runTestsByPath src/__tests__/db/queries.test.ts src/__tests__/db/queries-focus.test.ts` gave `3 failed, 47 passed`. Post-fix: `50 passed`.
- **Sabotage check — do the assertions still bite?** With rollback deliberately broken (rethrow preserved), 8 tests fail, every one of them on a row-count/state assertion (`Expected: 0 / Received: 1`, `Expected: 5 / Received: 6`), and "Received function did not throw" appears zero times. The tests still catch a genuine rollback regression, and now report it accurately.
- **Second sabotage — is the wrapper load-bearing?** Making `toRealmError` return a plain object brings back 13 failures with "did not throw", confirming the normalisation is doing real work and not swallowing rejections.
- **Flake soak.** Full suite, 10 consecutive runs: `757 passed` every time. Plus 5 runs each of `--runInBand` and `--runInBand --no-cache`, all `757 passed`. Re-soaked after later commits in an isolated clone: 20/20 db-suite runs green with parallel workers.
- `npx tsc --noEmit` clean. `npx eslint . --ext .ts,.tsx` 0 errors (10 pre-existing warnings on untouched files).

## Changes after the first review round

The numbers above are from the original implementation. Four review passes then landed three more commits — final suite is `759 passed`.

- **A failing `ROLLBACK` no longer discards the transaction error.** It previously replaced `err`, losing the root cause. Now caught, normalised, and the original attached as `cause`. This branch also gained a test — it was the subtlest logic in the file and had none.
- **The rollback assertion could have gone vacuous.** If a NOT NULL column were added to `entry`, the setup INSERT would fail and the test would pass for the wrong reason — a trap that already bit once during implementation. The in-transaction row count is now captured and asserted *outside* the transaction body; asserting inside gets swallowed by the capture helper and silently satisfies the assertions it guards.
- **The rollback test accepted any error** — swapping in a nonexistent table still passed it. Now pinned to `/FOREIGN KEY/`.
- A `/simplify` pass removed a `ROLLBACK` carve-out so all adapter throws normalise uniformly, collapsed duplicate assertions into `test.each`, and documented the raw-handle hazard beside `openTestDb` in `src/lib/db/test-helpers.ts` — the invariant keeping this fix complete was previously written down only in the one file where it cannot be violated.

One review finding was withdrawn: a report of rows surviving rollback turned out to be contamination from parallel critics sabotaging the same working tree, reproduced and confirmed as such in an isolated clone. Detail in the thread.

Closes #179

## Checklist
- [x] Acceptance criteria met
- [x] Flow tests (Maestro) added or updated for any user-facing behavior — N/A, test-infrastructure only
- [x] Migration test written if schema changed — N/A, no schema change
- [x] CI passing


sfrankle (commented):
## Simplify Pass

Four cleanup agents (reuse, simplification, efficiency, altitude). Findings deduped and applied in 2d30643.

### Applied

- **`ROLLBACK` carve-out removed** (`adapter.ts`). The file states "every method funnels thrown values through `toRealmError`", then exempted `ROLLBACK` as "genuinely exceptional, should surface as itself". But "itself" is exactly the shape Jest cannot classify — if that branch ever fired inside `rejects.toThrow()`, it would report "did not throw", the very failure #179 exists to remove, in the hardest condition in the file to debug. Now wrapped like everything else.
- **Redundant `rethrow` around `COMMIT` dropped.** It sat inside the `try` whose `catch` already normalises, so anything `COMMIT` threw was wrapped twice. Idempotent, so no behaviour change — just indirection that invited the reader to ask why `COMMIT` was special when `fn()` above it was not.
- **Three near-identical invalid-SQL tests collapsed into one `test.each`**, plus a shared `expectRealmLocalError` helper replacing five copies of the same two-line assertion pair. The contract now lives in one place instead of five.
- **`capture` widened to accept sync calls**, so the raw-error test uses it instead of hand-rolling a second try/catch with its own failure message.
- **Test-file header trimmed to the contract plus a pointer.** The realm/vm-context mechanism was narrated at length in both `adapter.ts` and the test file; two copies of a subtle explanation drift, and the one a reader opens first is the one that misleads. `adapter.ts` keeps it — it sits on the code being defended.
- **Hazard documented at the distribution point** (`test-helpers.ts`). `openTestDb()` hands raw better-sqlite3 handles to eleven test files, and the property keeping this fix complete — no raw handle awaited behind a `rejects` assertion — was written down only in the one file where it cannot be violated. Now noted beside `openTestDb`, so a future async integrity test announces the problem instead of reintroducing the flake elsewhere.

### Skipped

- **`{ cause: err }` on the wrapped Error** — flagged as unasserted, but it preserves the original `SqliteError` for debugging at near-zero cost. Keeping it.
- **Folding the `name` test into the FK test** — they cover different error classes (`no_such_table` vs a FK violation). Merging would trade a real distinction for one fewer block.
- **Replacing `rethrow` with inline try/catch** — benchmarked at 200k iterations: −56 ns/op, inside run-to-run noise, against a ~2.6 µs `prepare()` call. Three orders of magnitude cheaper than the work it wraps; inlining would duplicate the catch body four times for nothing.

### Verification

Re-ran the sabotage after refactoring, since a refactor of a guard test is exactly where a guard quietly stops guarding: stripping `toRealmError` fails 5 of the 8 guard tests. Full suite `757 passed`, `tsc --noEmit` clean, eslint clean on all three touched files.

sfrankle (commented):
## Safety Review

**Verdict:** PASS

### Privacy: ✅
No `fetch`/`axios`/HTTP client, no network permission, no analytics/crash/telemetry import, no new dependency. `package.json` is untouched. The only `expo-` occurrences in the diff are prose in comments and a decisions row describing what the test double stands in for. The only URLs added are `github.com/sfrankle/haven-app/pull/...` links in `docs/changelog.md`. No clinical or community terms anywhere in the diff, the six commit messages, the PR body, the changelog row, or the decisions row.

### Data Safety: ✅
No schema change and no migration — the five changed files are two test files, one test helper (comment-only change), and two docs. No `CREATE TABLE` / `ALTER TABLE` / `DROP` / `INSERT OR REPLACE` in the diff, so no migration test is owed.

The substantive risk here was the test double, since every db query test runs through it. Verified rather than taken on trust:

- **The normalisation cannot swallow a rejection.** `rethrow` has exactly one non-throwing path (`return fn()`); every catch ends in `throw`. `toRealmError` always returns a constructed `Error` — there is no branch that returns `undefined` or resolves.
- **`withTransactionAsync` still rolls back and still rejects.** `ROLLBACK` runs before `throw toRealmError(err)`, and an `Error` thrown by the transaction body passes through by identity (guarded by the `caught).toBe(original)` test).
- **Sabotage check, run independently.** Replacing `ROLLBACK` with `COMMIT` in the adapter across `adapter-errors`, `queries`, `queries-focus`, and `queries-routines`: **8 failed / 109 passed**, every failure on a row-count or row-state assertion (`Expected: 0 / Received: 1`, `Expected: 5 / Received: 6`), and `"did not throw"` appears **zero** times. A genuine rollback regression is still caught, and now reports accurately instead of as a phantom no-throw. Adapter restored afterwards; working tree clean.
- Unsabotaged, the same files are green (58 passed).

One non-blocking note: if `ROLLBACK` itself throws, its error replaces `err` and the original failure is discarded. The rejection is still preserved so nothing is masked, and the code comment calls this out — but passing the original as `{ cause: err }` there would keep the root cause of a failed transaction visible.

### Judgmental Language: ✅
One grep hit, `behind`, in an internal code comment in `src/__tests__/db/adapter-errors.test.ts` ("the mechanism behind the rule"). Not user-facing — no UI strings, error messages, or empty states are touched by this PR. No other terms from the list appear in added lines.

### Workflow Artifacts: ✅
- PR body contains `Closes #179`. `#179` is a standalone `technical-task` with no parent user story, so there is no `Contributes to #M` to add.
- `docs/changelog.md`: exactly 1 new row, using the PR number (181, not the issue number), as a markdown link to the full PR URL, written for a skimmer with no implementation detail.
- `docs/decisions.md`: 1 row appended at the **end** of the table (`@@ -47,3 +47,4 @@`, added as the final line) — not inserted mid-table.
- `CI passing` checkbox is correctly **not** pre-checked. (For reference, Type-check/Lint/Test is currently green.)
- No milestone — correct, `#179` has none.

All safety checks passed.

sfrankle (commented):
## Product Vision Review

**Verdict:** PASS

Scoping note up front: #179 has no parent user story and no milestone, and this PR touches no production code, UI, copy, or schema. There is no user-facing surface to evaluate. The honest product-vision angle here is narrow — it is **data-safety confidence**, and nothing else. I have not manufactured UX findings where none apply.

### Data-safety posture (the part that actually matters)

The ticket's argument is the right one, and it is a Haven-principles argument rather than a hygiene one. `docs/data/schema.md` rollback behaviour is the guarantee that a user's log never lands half-written. A test that goes red for a reason unrelated to rollback makes the red signal ambiguous, and an ambiguous red signal is how a genuine rollback regression gets waved through. Fixing the reporting so that a rollback failure and a realm mismatch are no longer indistinguishable is a direct strengthening of the highest-stakes concern in the codebase.

The diagnosis is also the correct *kind* of fix. The adapter is a double standing in for expo-sqlite, which rejects with genuine realm-local `Error`s; better-sqlite3's `SqliteError` does not. Normalising at the double's boundary closes a fidelity gap between the double and the real thing. That is the opposite of loosening a test — the double now lies less about production.

### Acceptance criteria — independently verified

I re-ran all three rather than taking the PR description on trust.

1. **Root cause identified and documented** — met. The mechanism (`Error.call(this)` instead of `super()`, missing `[[ErrorData]]` slot, process-wide addon binding vs. per-file vm contexts, Jest's `isError` gate in `toThrowMatchers`) is documented in three places at the right level of detail for each: the `adapter.ts` header for the person editing it, `test-helpers.ts` for the person reaching for a raw handle, and one `docs/decisions.md` row for the durable record. Alternatives considered are recorded honestly, including the ones that would have been weakenings.

2. **Full suite passes across 10 consecutive runs** — met. 10/10 runs, `757 passed` every time.

   Worth recording how I got there: partway through my first soak the working tree was switched to `main` by another process, so several of my runs silently executed *without* this fix. Those runs failed — `query layer › saveEntry › transaction rollback on failure — non-existent labelId commits no entry row`, reported as "Received function did not throw". That was an accident, but it is a clean independent reproduction of the bug on `main`, and it confirms both the failure mode and the misleading error text described in #179. I then re-ran the soak in an isolated clone pinned to `2d30643` and got 10/10 clean.

3. **No test weakened or skipped; rollback assertions still bite** — met, and this is the criterion I most wanted to see fail. I deliberately broke rollback (swapped `ROLLBACK` for `COMMIT` in `withTransactionAsync`) and ran the db suites: **7 tests failed, every single one on a row-count or row-state assertion** (`Expected: 0 / Received: 1`, `Expected: 5 / Received: 6`, `Expected: "Existing" / Received: "Good item"`). "Received function did not throw" appeared **zero** times. The assertions still catch a real rollback regression, and they now name it accurately instead of pointing at the wrong thing. Restored cleanly afterwards.

The choice to guard on `Object.prototype.toString.call(err) === '[object Error]'` rather than `instanceof Error` is the load-bearing detail and it is reasoned correctly — `instanceof` is the exact value that varies with load order, so depending on it would let the first-loaded file skip wrapping and reintroduce the asymmetry. `adapter-errors.test.ts` locks the contract, and its first test is nicely self-retiring: if better-sqlite3 ever makes `SqliteError` a real Error exotic object, that assertion flips and tells you the normalisation can go.

### Documentation voice

**Changelog row** — good. "Fixed an intermittent test-suite failure that made database rollback tests look broken when they weren't" is one sentence, plain, and leads with the consequence rather than the mechanism. No `SqliteError`, no realms, no `[[ErrorData]]`. A skimmer six months from now learns the useful thing: the tests were misreporting, the data layer was fine. It sits comfortably next to the other internal row (#177) without pretending to be a user-facing feature. No blame framing, nothing alarming.

**decisions.md row** — correct content, correct columns, appended at the end as required. Voice is fine for an internal engineering doc; the tone rules in `docs/design/brand.md` govern user-facing copy and don't bind here. It is noticeably the longest row in the table, but the density is earned — this is a subtle, load-order-dependent failure and an under-documented row would strand the next person.

### UX principles / tone / interaction

Not applicable. No user-facing strings, no empty or error states, no flows, no spacing or rhythm decisions. Nothing to assess.

### Side effects

None found. The blast radius is three test-support files plus two doc rows. No jest config, no CI config, no `setupFilesAfterEach` hook, no production module. Nothing changes what a user sees or what is written to their database. The 2000-iteration probe confirming rollback was never actually broken is the right instinct — it establishes that no user data was ever at risk, which was the open question the flake left hanging.

### Residual risk (noting, not blocking)

The normalisation lives at the `createAdapter` boundary, so a future test that awaits a raw `openTestDb()` handle and asserts `rejects.toThrow()` would bring the flake back. This is called out in both `test-helpers.ts` and the decisions row, which is the right minimum, but it currently relies on someone reading the comment. If this recurs, a lint rule or an ESLint `no-restricted-syntax` guard would be the natural follow-up — not needed now.

Separately: if `ROLLBACK` itself throws, its error replaces the original `err` and the root cause is lost. The inline comment acknowledges this and the priority ordering is defensible (a failed rollback is the more urgent fact), so I'd leave it.

Implementation aligns with Haven's product vision — specifically, it makes the data-safety signal trustworthy again, which is the highest-stakes thing this codebase has.

sfrankle (commented):
## Code Quality Review

**Verdict:** CONCERNS

Code itself is clean — `tsc --noEmit` and `eslint` both pass, no `any`, no dead code, no debug statements, decisions.md row appended at the end, changelog row at the top with a link. The plan-critic items all held: the guard is shape-based (`Object.prototype.toString`) not `instanceof`, all four adapter methods wrap, and the previously-vacuous rollback assertion now has a valid setup INSERT (`entry` has exactly four NOT NULL/no-default columns and all four are supplied).

The concern is empirical, not stylistic.

### The suite is still intermittently red at head

Running `npx jest src/__tests__/db --ci` repeatedly at `2d30643`, I got **5 red runs out of the first ~14**, each with the same 8 failures across 4 files:

- `adapter-errors.test.ts` › withTransactionAsync rejects with a realm-local Error and rolls back
- `queries.test.ts` › saveEntry › transaction rollback on failure
- `queries-focus.test.ts` › (2 rollback tests)
- `queries-routines.test.ts` › (4 rollback tests)

Critically, the **failure mode has changed**. At base (`872a586`) every run was red with `Received function did not throw` — the realm problem this PR fixes. At head the `toThrow`/`expectRealmLocalError` assertions **pass** and the test proceeds to the row-count assertion, which then fails:

```
● adapter error normalisation › withTransactionAsync rejects with a realm-local Error and rolls back
  expect(received).toBe(expected)
  Expected: 0
  Received: 1
  > 113 |     expect(countAfter).toBe(countBefore);
```

The error was thrown, it was realm-local, `raw.inTransaction` was `false` — and the row survived. For a row to survive, the INSERT must have been auto-committed, i.e. `BEGIN` did not take effect. That is the *genuine rollback regression* #179 said was indistinguishable from the realm mismatch: the fix has made the two distinguishable and, in those runs, it was the genuine one.

Two caveats, stated plainly: the failures only occur with parallel workers (`--runInBand` is green 3/3), and after roughly the first 15 runs I could not reproduce it again in ~50 further attempts (varying `-w`, cold cache, heavy CPU load, concurrent `tsc`/`eslint`). So the window has clearly narrowed a great deal — but "narrower" is not "closed", and the residual mode is the dangerous one.

**Ask:** before closing #179, run `npm test -- --ci` (default workers, not `--runInBand`) 20+ consecutive times in CI and paste the result. If any run is red, the residual belongs in a follow-up issue framed as *transaction semantics under jest's per-context module loading*, not as error reporting. Also worth softening the decisions.md claim that normalising "closes a fidelity gap rather than weakening a test" — that is true of the error typing, but it is not yet established that no genuine rollback problem exists underneath.

### Tests

- `adapter-errors.test.ts:98` — the comment `// Violates the FK on label_id — 999999 does not exist` is wrong. The args are `[999999, labelId]`, so `entry_id` is the bogus one and `label_id` is valid. Fix the comment (or swap the args to match it).
- `adapter-errors.test.ts:113` — `expect(countAfter).toBe(countBefore)` is non-vacuous today but silently *becomes* vacuous if anyone adds a NOT NULL column to `entry`: it also passes when the setup INSERT never landed. Cheap insurance is to assert inside the transaction body, before the failing insert, that the count is `countBefore + 1`. That pins the thing the test is actually about.
- `toRealmError`'s non-object path is untested. `throw 'a string'` / `throw null` hit the `String(err)` branch and nothing covers it; a `cause` preservation assertion is similarly free to add.
- Otherwise test naming and structure are good — names describe behaviour, `test.each` + `expectRealmLocalError` read well, and the "passes through by identity" test is the right way to pin non-wrapping.

### TypeScript

- `expectRealmLocalError(err: unknown)` has no explicit return type; add `: void` for consistency with `capture`.
- Casts in the test file (`as Error & { code?: string }`) are appropriate for assertion narrowing, not hiding errors.

### RN/Expo, Haven patterns

No production code, no schema, no migration, no navigation or UI surface — nothing to flag. decisions.md entry is thorough and correctly placed.

### General

Clean. No console statements, no commented-out code, no leftovers.

sfrankle (commented):
## Code Quality Review (re-run @ d784aee)

**Verdict:** PASS

All four findings from the previous review are correctly fixed. Verified in an isolated clone at `d784aee` (`node_modules` symlinked, main working tree untouched) — 6/6 `src/__tests__/db` runs green with parallel workers, `tsc --noEmit` clean, `eslint` clean on touched files. I no longer see the intermittent red I reported previously; the contamination analysis holds.

### Verification of the four fixes

**1. `cause` chaining on a failing ROLLBACK — correct.** Exercised it directly in the clone (run `ROLLBACK` inside the body, then throw): the thrown value is `cannot rollback - no transaction is active` with `cause` === the original body error, identity-preserved.

**2. Comment corrected to `entry_id`** — matches the bogus argument. Confirmed.

**3. Non-vacuity guard — your reasoning is right, and I verified it both ways.** Breaking the first INSERT makes the test fail with `Expected: 1 / Received: -1` at the `countInside` assertion. Moving that same assertion back *inside* the body with the same sabotage makes the test pass green — swallowed by `capture()`, exactly as your comment says. Asserting outside is the correct call.

**4. `expectRealmLocalError(): void`** plus the new `cause` survival test. Confirmed.

### Findings

**Tests (minor, optional)**
- The rollback test is now non-vacuous on the *first* INSERT but not on the *second*. I swapped `entry_label` for a nonexistent table and the test still passed green — any error satisfies it. If the FK on `entry_label.entry_id` is ever dropped or renamed, the test keeps passing for the wrong reason. One line closes the symmetric hole: `expect((err as Error).message).toMatch(/FOREIGN KEY/)`.
- The failing-ROLLBACK branch is the newest and subtlest logic in the file and has no test. It is not contrived to reach — an inner `await db.runAsync('ROLLBACK')` followed by a `throw` hits it in ~8 lines and asserts `caught.cause === original`. Worth pinning, since a regression here is silent.
- Agreed on leaving `toRealmError`'s `String(err)` branch untested — unreachable through the adapter, and `toRealmError` isn't exported, so testing it would mean widening the API for a fake. Not worth forcing.

**General (nit)**
- In the ROLLBACK-failure path, when `rollbackErr` is a `SqliteError`, `toRealmError` sets `cause: rollbackErr` and the next line overwrites it with the transaction error. `message`/`name`/`code` survive on the wrapper, so nothing important is lost, but the raw rollback object is dropped — slightly at odds with the contract the `cause` test documents ("the untouched original is kept on `cause`"). A one-line comment saying the transaction error deliberately wins the `cause` slot would remove the ambiguity. Same line also clobbers any pre-existing `cause` if `rollbackErr` arrives as an already-realm-local `Error`; edge enough not to matter.

**RN/Expo, TypeScript, Haven Patterns:** no issues. Test-double-boundary casts only, `decisions.md` entry present and unusually thorough (including the alternatives considered and the conditions under which the flake returns), changelog row present. No runtime/UI surface, so no Maestro impact.

Code quality looks good — ship it; the two test suggestions are non-blocking.

sfrankle (commented):
## Post-critic fixes

Three critics ran, then one re-ran. Final: code-quality **PASS** (re-run), product-vision **PASS**, safety **PASS**.

### One finding was mine, not the code's

The code-quality critic's first pass reported 5 of ~14 runs red at head, with rows surviving a rollback — a genuine data-loss-shaped failure. It was contamination I caused: I dispatched all three critics in parallel into one working tree, and two of them apply sabotage edits (`ROLLBACK` → `COMMIT`) to verify the assertions still bite. The third was soak-testing while those edits were live.

Two tells: it could not reproduce in ~50 later attempts, and the product-vision critic independently noticed "another process switched the working tree to main" mid-soak.

Verified in an isolated clone pinned to the head SHA, `node_modules` symlinked:

- 20/20 db-suite runs green with parallel workers
- deliberate sabotage in that clone reproduced the reported signature exactly — `8 failed`, `Expected: 0 / Received: 1`, zero "did not throw"

The re-run, told to isolate, independently could not reproduce the red either. Rollback is fine. Noted so the next reader of this thread does not treat the first report as an open question.

### Fixes applied (d784aee, 7d02435)

- **`ROLLBACK` failure no longer discards the transaction error.** Previously a throwing ROLLBACK replaced `err` and the root cause vanished. It is now caught, normalised, and the original attached as `cause`. Flagged by all three critics.
- **The rollback assertion could go vacuous.** If a NOT NULL column were added to `entry`, the first INSERT would fail and the rollback test would pass for the wrong reason — the same trap that already bit once during implementation. The in-transaction count is now captured and asserted *outside* the body; asserting inside would be swallowed by `capture()` and silently satisfy the assertions it guards. The critic verified both directions: sabotaging the INSERT fails it at `Received: -1`, and moving that assertion back inside makes the same sabotage pass green.
- **The rollback test accepted any error.** Replacing `entry_label` with a nonexistent table still passed it. Now pinned to `/FOREIGN KEY/`.
- **The failing-ROLLBACK branch had no test** — the subtlest logic in the file. Now covered.
- Corrected a comment naming the wrong column (`entry_id`, not `label_id`); added `: void`; added a test that `cause` survives normalisation.

### Left alone

`toRealmError`'s non-object `String(err)` branch stays untested — reaching it means exporting the helper or contriving a fake, and widening the API to test an unreachable branch is a worse trade. Both the critic and I landed there independently.

Full suite `759 passed`, `tsc --noEmit` clean, eslint clean on touched files.

Issue #126, opened 2026-04-08
Routine grouping in Trace


sfrankle:
## Story
As a user, I want entries from a single Routine completion to appear as one grouped row in Trace so that completing a Routine 3x a day doesn't flood my history with individual entries.

## Acceptance criteria
- [ ] Entries sharing a `routine_completion_id` are displayed as a single collapsed row in Trace
- [ ] Collapsed row shows: Routine name + timestamp of completion
- [ ] Tapping the row expands it to show all individual items with their logged values (including any notes)
- [ ] Grouping is deterministic — based on `routine_completion_id`, not heuristic
- [ ] When a Focus filter is active, a grouped row is shown if any entry in the group matches the Focus

PR #182, opened 2026-07-28
Routine grouping in Trace (multi-filter support)

### LINKED ISSUE #165: Routine grouping in Trace (multi-filter support)
## Summary
Group Routine completion entries in Trace into a single collapsed row per `routine_completion_id`, with expand/collapse and correct behaviour under any combination of active filters. Also upgrades the Trace filter mechanism from single-Focus to multi-filter.

### Multi-filter upgrade
- [ ] Trace filter state changes from a single `activeFocusId` to a set of active filter IDs (Focus and/or other future filter types)
- [ ] Multiple Focus pills can be active simultaneously; an entry matches if it belongs to any active Focus
- [ ] Single-pill selection behaviour is preserved as a subset of the new model

### Routine grouping
- [ ] `getEntriesForTrace` (or a companion query) returns entries with `routine_completion_id` populated where applicable
- [ ] Entries sharing a `routine_completion_id` are rendered as a single collapsed `RoutineGroupRow`: shows Routine name + completion timestamp
- [ ] Tapping `RoutineGroupRow` expands to show all items with their logged values (labels, numeric value, notes)
- [ ] Grouping is deterministic — based on `routine_completion_id`, never heuristic
- [ ] Under active filters: a group row is shown if any entry in the group matches any active filter; expanded view renders matching entries at full opacity, non-matching entries muted (opacity 0.45, matching the existing Focus-filter muting pattern)
- [ ] Tests: unit tests for grouping logic, multi-filter entry matching, muting behaviour

## Notes
The `getEntriesForTrace` function currently has two query shapes (unfiltered, filtered-by-focus) following the two-constant pattern from the decisions log. Adding routine grouping is a third concern — consider whether to extend the existing function or post-process at the hook level. Whichever approach is chosen, document the decision in `docs/decisions.md`.

Blocked by: #164

sfrankle:
## Summary
- Routine completions now read as one row in Trace instead of a scatter of individual entries. Entries sharing a `routine_completion_id` collapse into a single group row (Routine name + completion time) that expands to show every logged item with its labels, values, and notes.
- Trace filtering goes from one-at-a-time to multi-select: `selectedFocusId` is replaced by `activeFilters: TraceFilter[]`, so several Focus pills can be active at once and an entry matches if it belongs to *any* of them. Single-pill selection still behaves exactly as before — it's just the one-element case.
- Read-path and presentation only. No schema change, no migration; `routine_completion_id` already exists on `entry` from the v9 migration.

Closes #165
Contributes to #126

## What to review
- **`src/lib/db/queries.ts` — `WHERE EXISTS` semi-join.** Multi-filter is applied with a correlated `EXISTS` subquery rather than widening the existing JOIN. A widening JOIN returns one row per matching focus/label, and `collapseTraceRows` would then emit duplicate label chips for an entry matching two active Focuses. The semi-join keeps the result set one-row-per-entry-label, which is what the collapse step already assumes.
- **`src/lib/utils/traceUtils.ts` — `buildTraceItems`.** Grouping happens in JS, not SQL. `getEntriesForTrace` stays a flat entry read; a companion `getRoutineCompletions` query supplies the Routine name and completion timestamp, and `buildTraceItems` folds entries into group rows by `routine_completion_id`. This keeps the existing two-query-shape pattern in `getEntriesForTrace` from growing a third shape. Rationale recorded in `docs/decisions.md`.
- **`src/hooks/useTraceEntries.ts` — degradation path.** If `getRoutineCompletions` fails, the hook silently falls back to ungrouped entries rather than surfacing an error. Losing the grouping affordance is a much smaller failure than blanking a user's history, and every entry is still individually visible.
- **`app/(tabs)/trace.tsx`** — group row expand/collapse, and the muting rule under active filters: a group row appears if any entry in it matches, and when expanded the non-matching entries render at opacity 0.45, reusing the existing Focus-filter muting pattern.
- **`src/lib/utils/traceFilters.ts`** — the `TraceFilter` union and toggle/select helpers. The union is deliberately open so future non-Focus filter types slot in without another state refactor.

## Testing
- `npx tsc --noEmit` — clean.
- `npx eslint . --ext .ts,.tsx` — 0 errors, 10 pre-existing warnings.
- `npm test -- --ci` — 70 suites, 824 tests, all passing.
- New/extended unit coverage: `getEntriesForTrace` multi-filter semi-join, `getRoutineCompletions`, `buildTraceItems` grouping and ordering, `traceFilters` toggle/select helpers, `useTraceEntries` orchestration including the `getRoutineCompletions` failure fallback, and `TraceScreen` rendering for group rows, expand/collapse, and non-matching-entry muting. Every acceptance criterion on #165 is carried by a Jest test.
- **Maestro flow `maestro/flows/trace/routine-grouping.yaml` was added but not run** — no simulator was available in this environment. Maestro is not part of CI, so this flow is unverified and should be run manually (or on next device pass) before relying on it.

## Checklist
- [x] Acceptance criteria met
- [ ] Migration test written if schema changed — N/A, no schema change
- [x] CI passing


sfrankle (commented):
## Simplify Pass

Four cleanup reviews (reuse, simplification, efficiency, altitude) ran over the branch diff. Fixes applied in `390e91d`; `npx tsc --noEmit` clean, eslint 0 errors (10 pre-existing warnings), 823/823 Jest tests passing.

### Fixed

- **Extracted `EntryDetail`** — `RoutineGroupRow` had copy-pasted `EntryRow`'s expanded body (chips + notes guard). The `formatTraceChipLabel` call carries the Physical severity/area rules, so the two copies were already one edit away from diverging. Now one component, testIDs parameterised.
- **Extracted `toggleInSet`** — `handleToggleGroup` was a verbatim copy of `handleToggle` ten lines below it.
- **`EntryIcon` in the group row** — was a hand-rolled `MaterialCommunityIcons` with the same size/colour, so a theme change to `EntryIcon` would have skipped Routine rows.
- **Dropped the serialize/parse round-trip in `useTraceEntries`** — the effect joined `focusIds` to a string and then `.split(',').map(Number)`'d it back to reconstruct a value already in scope. The effect now reads `focusIds` directly; `focusKey` stays as the stable dep. Paired with a `useMemo` on `focusIdsOf(activeFilters)` at the call site so the array identity is stable rather than laundered.
- **Empty-state copy and context back label now key off `activeFilters.length`, not `focusIds.length`** — two competing definitions of "is a filter active" in one component. The first non-focus filter variant would have shown "Nothing logged yet." for a filtered-empty list.
- **Removed `orderedCompletionIds`** in `getRoutineCompletions` — a `Map` already iterates in insertion order, so it duplicated the Map's own ordering plus a non-null assertion.
- **Merged the two identical fallback branches** in `buildTraceItems`, so "degrade to ungrouped" has one exit point.
- **`getRoutineDayProgress` now calls the new `placeholders()` helper** — its inline copy shadowed the module function.
- **`keyExtractor` → module-scope `traceItemKey`** — `useCallback` with empty deps for a pure function.

### Raised and deliberately skipped

- **`getRoutineCompletions` re-implements `collapseTraceRows`** — a second collapser over the same `EntryTraceRaw` shape, which is exactly the drift the file's own doc comment warns about. Real, but reworking it to bucket via a rowId→completionId map is a behaviour-carrying change, not a cleanup. Worth a follow-up.
- **`matchedIds` on group items mixes list-wide filter state into the render model.** Flagged by two reviewers; the alternative (mark `matched` per entry at build time) is a `TraceItem` data-model change with test churn.
- **Lazy-loading group members / metadata-only query when unfiltered.** Genuine wasted I/O — the unfiltered case re-fetches entries it already has — but it restructures the read path the plan deliberately chose.
- **`renderItem` identity churn** — expanding one group re-renders every visible row. Fixing it properly means `React.memo` + `extraData`, a perf refactor beyond this pass.
- **Collapsing the `TraceFilter` union to `number[]`** — suggested as over-abstraction, but the extension point is an explicit acceptance criterion on #165.
- **Shared test fixtures** for the three private `makeDb()` / `EntryTraceRaw` literals — test-only, and this PR already had to touch all three.

sfrankle (commented):
## Safety Review

**Verdict:** PASS

### Privacy: ✅
No `fetch`, `axios`, `XMLHttpRequest`, or WebSocket calls introduced. No analytics, crash-reporting, or telemetry imports. No new dependencies at all — `package.json` is untouched, so no new `expo-*` package and no new permission. The only URL anywhere in the diff is the GitHub PR link in the `docs/changelog.md` row. Read path stays entirely on-device.

### Data Safety: ✅
The "no schema change, no migration" claim is confirmed against the diff. `src/lib/db/migrations/` has no added or modified files (latest remains `v9__routine-tables.sql`), and `routine_completion_id` already exists on `entry` on `main` via v9 — so no migration is owed and the N/A on the migration checklist item is correct. No `CREATE TABLE`, `ALTER TABLE`, or `DROP` in non-test code. Every `INSERT` in the diff is test-fixture setup inside `src/__tests__/` and `src/lib/db/__tests__/`; no seed data is added, so the `INSERT OR IGNORE` rule doesn't apply here. Nothing drops a column or table, and no write path is modified — this PR is read and presentation only.

**SQL injection check on `placeholders(n)` — clean.** The helper is `Array.from({ length: n }, () => '?').join(', ')`; it consumes only a number and can only ever emit `?` characters, so no caller-supplied value can reach the SQL string. All three call sites pass a `.length`, never a value:
- `buildFilteredTraceSql(focusIds.length)` → IDs bound via `getAllAsync(sql, focusIds)`
- `getRoutineCompletions`: `placeholders(completionIds.length)` → bound via `completionIds`
- `getRoutineDayProgress`: `placeholders(routineIds.length)` → bound via `[...routineIds, today]`

The empty-list guard is right in both variable-length queries: `getEntriesForTrace` routes `focusIds.length === 0` to the unfiltered SQL and `getRoutineCompletions` early-returns `[]`, so neither can emit `IN ()`. The `WHERE EXISTS` semi-join also correctly avoids the row-multiplication that would duplicate label chips through `collapseTraceRows`.

### Judgmental Language: ✅
No matches in user-facing code. New strings in `src/constants/messages.ts` are "Nothing logged yet.", "No entries for this Focus yet.", "No entries for these Focuses yet." — all neutral, no negative framing on the empty state. The banned-term grep hits only non-user-facing lines: source comments ("never blank the user's history behind an error"), a test error fixture (`new Error('trace read failed')`), and a test comment. No "streak", "score", "rank", or "incomplete" anywhere.

### Workflow Artifacts: ✅
- `Closes #165` (technical task) present
- `Contributes to #126` (user story) present
- `docs/changelog.md`: exactly one new row, PR number rendered as a markdown link to the full PR URL, uses **#182** (the PR number) rather than the issue number, appended at the top per convention
- `CI passing` checkbox is correctly left **unchecked**
- `docs/decisions.md` rows appended at the end of the table, not inserted mid-table

All safety checks passed.

One non-blocking note for the human reviewer, outside my binary criteria: the PR body states the Maestro flow `maestro/flows/trace/routine-grouping.yaml` was added but never run, while the "Flow tests added or updated" box is checked. That's honestly disclosed and Maestro isn't in CI, so it isn't a safety failure — just worth a manual device pass before merge.

sfrankle (commented):
## Product Vision Review

**Verdict:** CONCERNS (none blocking — all findings are polish or follow-up)

Reviewed against user story #126 and technical task #165.

### User Story Fulfillment

All six acceptance criteria on #126 are met in spirit, not just technically:
- Grouping is keyed strictly off `routine_completion_id` (`buildTraceItems`), never timestamp proximity — deterministic as required.
- Collapsed row shows Routine name + completion time; expanding shows every member with labels, numeric values, and notes via the shared `EntryDetail`.
- Under an active filter a group surfaces if any member matched, and non-matching members render at 0.45 opacity — the same muting vocabulary already used by the Focus filter and the context view, so it reads as one idea rather than a new one.

The core user problem — "completing a Routine 3x a day floods my history" — is genuinely solved. Three completions now read as three rows.

Two robustness behaviours are also good product calls, not just good engineering:
- A completion whose `routine` row is gone (or whose enrichment query failed) falls back to individual entries. **Nothing a user logged ever disappears from Trace** — the worst case is losing the collapse, which is the right direction to fail in.
- A one-item Routine still renders as a group row rather than switching presentation at count 2, so the list doesn't shape-shift as a Routine grows.

### The two approved UX decisions

Both hold up.

**No "Show context" on group rows** — correct. A ±2h window has to be anchored somewhere, and a group has no single anchor. Offering it would have meant inventing a rule the user can't see.

*Follow-up worth considering (not for this PR):* the consequence is that entries logged through a Routine now have **no** path to context at all — not even individually, because expanded members aren't interactive. Each member does have exactly one unambiguous timestamp, so surfacing "Show context" on an expanded member would close that gap later without touching the group row rule.

**No "1 of 4 match" suffix** — agreed, that phrasing counts things at the user and edges toward scoring. The muting-on-expand is the honest explanation and it arrives at the moment the user asks for it.

One residual: a collapsed group row today carries no visible signal that it contains more than one thing (the count exists only in the accessibility label). Under a filter, "Trace Group Test · 08:12" can read as "this whole Routine matched." A neutral descriptive count on the row — e.g. `Morning Flow · 4 items` — would resolve both that ambiguity and the general "is this one thing or several?" question, without the comparative framing that made "1 of 4 match" wrong. Suggestion, not a requirement.

### UX Principles

No scores, no streaks, no completion percentages, no ranking. The Routine group row describes what happened and stops there — it does not evaluate the Routine. Correct for Haven.

Empty states are present for all three filter cardinalities and none frame absence as failure. Error state is unchanged and neutral.

**Degradation path (per the plan critic's concern) — verified from the user's side.** `useTraceEntries` catches the `getRoutineCompletions` failure and passes `groups: []` into `buildTraceItems`, which emits every entry individually. Trace never blanks, never shows an error screen, and no entry is hidden. The failure is invisible to the user apart from the missing collapse. This is exactly the right trade and the code comment makes it hard to "fix" by accident.

Minor: the PR description says the hook "logs and falls back", but the `catch {}` is silent — there is no log. Not user-facing; either add the log or correct the description.

### Tone and Copy

`src/constants/messages.ts` additions all pass:
- `traceEmpty: "Nothing logged yet."` — neutral, factual, forward-looking. No implication the user should have logged something.
- `traceEmptyFiltered: "No entries for this Focus yet."` / `traceEmptyFilteredPlural: "No entries for these Focuses yet."` — describes the filter, not the user. The trailing "yet" keeps both open rather than terminal.

No instances of missed / failed / incomplete / overdue / behind anywhere in the diff. "Focuses" as a plural matches existing product usage in the changelog and decisions log.

The `ContextView` back label change ("Back to {name} filter" → "Back to filters" when several are active) is honest and stays plain-language; the single-filter wording is preserved byte-for-byte, so nobody's existing flow reads differently.

### Interaction Feel

One real issue. Inside an expanded group, each member is a bare `View` with only conditional styling for muting — there is no vertical gap between consecutive members. `ChipTray` contributes `marginTop: sectionGap` above its own chips, but nothing separates one member's notes from the next member's summary line. Two members with no labels and no notes will render as two flush lines of body text with zero space. Against `docs/design/interaction.md`'s spacing rhythm (12dp minor element gap), and against "calm by default" — a Routine's items should read as distinct logged things, not as a paragraph. Recommend a `spacing.elementGap` between members (and confirm the visual on a device before relying on the unrun Maestro flow).

Secondary, lower confidence without a device: group members render without the entry-type icon and without a per-item time, while every other row in Trace leads with an icon. The completion timestamp on the header justifies dropping the time, but the missing icon removes the scanning affordance that makes the rest of Trace fast to read. Worth eyeballing.

Expand/collapse itself matches the existing entry-row idiom exactly — tap to open, no chevron, no confirmation, no navigation. Group expansion state resets on screen focus and on tab re-press alongside entry expansion state, so there are no stale-open rows and no dead ends. Multi-select filtering is additive and reversible with no intermediate "apply" step, which fits Haven's rhythm.

### Side Effects

Nothing out of scope changed from the user's perspective:
- Single-Focus filtering is preserved as the one-element case; the semi-join means an entry in two active Focuses shows its labels once rather than twice, which is the behaviour a user would expect and would have been visibly wrong under a widening join.
- `getContextEntries` was refactored onto the shared SELECT but its output shape is unchanged and pinned by a test.
- Filter-reset-on-tab-press behaviour is unchanged.
- No schema change, no migration, no data written. Read and presentation only — no path by which existing logged data can be altered or lost.

Implementation serves the user story. The spacing inside expanded groups is the one thing I'd fix before this ships.

sfrankle (commented):
## Code Quality Review

**Verdict:** PASS

Reviewed `792750f..390e91d` (post-`/simplify`). `npx tsc --noEmit` clean, `eslint` 0 errors / 10 pre-existing warnings, `npm test -- --ci` 70 suites / 823 tests green.

### Plan-critic concerns — all five verified

1. **D1 regression test is real, not vacuous.** `src/__tests__/db/queries.test.ts` runs against the better-sqlite3 adapter with real migrations + seed. I verified non-vacuity by sabotaging `buildFilteredTraceSql` in an isolated clone — swapping the `WHERE EXISTS` semi-join for the widening `JOIN entry_focus ... IN (?,?)` fails `'an entry matching two active focuses is returned once, with its labels unduplicated'` with `Expected length: 3 / Received length: 6`. Exactly the predicted 2 focuses x 3 labels fan-out. This is the right test in the right file.
2. **Degradation path honoured.** `useTraceEntries` wraps only `getRoutineCompletions` in its own try/catch, falls back to `groups = []`, and `buildTraceItems` converges the enrichment-failure and unresolvable-completion-id cases onto one code path. Covered at both levels: `'degrades to an ungrouped list when getRoutineCompletions rejects'` (asserts `error` stays null and both entries still render) and `'renders every routine entry individually when groups is empty'`.
3. **`maestro/subflows/create-morning-routine.yaml` untouched.** `git diff --name-only maestro/` returns only the new `flows/trace/routine-grouping.yaml`, which builds its own two-item Routine and documents in its header exactly why it does not reuse the shared subflow.
4. **Jest carries the acceptance criteria.** Group row render/expand/collapse, muting under filter, absence of Show-context inside a group, multi-pill select/deselect, tab-retap clear, and both empty-state strings all have Jest coverage independent of Maestro.
5. **`getContextEntries.test.ts` pins the shared-shape hazard.** Two tests: SQL contains `e.routine_completion_id`, and the column maps onto the returned entry.

### Findings

**RN/Expo** — No issues. The `eslint-disable react-hooks/exhaustive-deps` in `useTraceEntries` is the one deviation and it is the correct call: `focusKey` is a total, order-insensitive proxy for `focusIds`, the comment explains why the captured array is never stale, and the `expo-router` mock was upgraded to `useEffect(cb, [cb])` so it re-runs on callback identity — which makes `'does not reload when re-rendered with a new array of the same contents'` a genuine guard rather than a mock artefact. Cancellation is correct across both awaits (`'does not set state after unmount between the two queries'`). `focusIds` is memoised at the call site. Two expansion sets rather than one avoids the entry-id/completion-id key collision, and there is a test for it.

**TypeScript** — No issues. `TraceItem` and `TraceFilter` are discriminated unions narrowed properly at every use. The only `as unknown as` is in `traceFilters.test.ts`, casting a future-shaped filter member so `focusIdsOf`'s narrowing is exercised instead of vacuous — commented and justified. All new components have props interfaces.

**Haven patterns** — Both new patterns have `docs/decisions.md` rows appended at the end of the table (semi-join + count-derived placeholder interpolation; JS-side grouping + enrichment-failure degradation). The `placeholders(n)` helper is derived solely from array length, never from a value, and `getRoutineDayProgress` was migrated onto it. `focusIds: []` correctly routes to the unfiltered SQL — `IN ()` is a SQLite syntax error, and there is a test pinning it. `getRoutineCompletions` joins `routine` via `routine_completion.routine_id` rather than the `ON DELETE SET NULL` `entry.routine_id`, with a test asserting the join text. Changelog row added, one sentence, linked.

**Tests** — Strong. Behaviour-level assertions throughout; the shared `TRACE_ENTRY_SELECT`/`collapseTraceRows` hazard is called out in a comment *and* pinned by tests. Good edge coverage: empty id list, single-member group, orphan completion id, non-adjacent members, wall-clock `localDate` slice vs UTC normalisation, and `completedAt` deliberately differing from every member timestamp so it cannot silently come from an entry.

### Minor (non-blocking)

- **PR description drift.** The body says "the hook logs and falls back"; the code is a bare `catch { groups = [] }` with no log. Either add a dev-only log or correct the description — right now an enrichment failure leaves no diagnostic trace at all.
- **`accessibilityLabel` does not pluralise.** `RoutineGroupRow` builds `${itemCount} items`, so a one-item Routine reads "1 items" to a screen reader. One-line fix.
- **Empty-state copy keys off `filterCount` alone.** `contextBackLabel` guards with `filterCount === 1 && focusIds.length === 1`, but `emptyMessage` only checks `filterCount`, so a future non-focus filter would render "No entries for this Focus yet." No effect today — `TraceFilter` has one member — but the two branches disagree about the same question.
- **Midnight section ordering is documented but untested.** `groupTraceItemsByDate` files a group under `group.localDate` while its stream position comes from its newest matching member, which can be a different day. I worked both directions through and the existing per-section `reverse()` recovers correct chronology in each; worth a test to keep it that way.

Code quality looks good.

sfrankle (commented):
## Product Vision Review — re-review after 9a429da

**Verdict:** PASS

### What I re-checked

**Expanded group member spacing (my one ship-blocker) — resolved.** The expanded group container now composes `styles.expandedGroupContainer` ({ gap: spacing.elementGap }) alongside `styles.expandedContainer`. `spacing.elementGap` resolves to 12 in `src/constants/theme/tokens.ts`, which is exactly interaction.md's canonical minor element gap — the fix uses the design token rather than a hand-picked number, so it stays in rhythm if the token moves. `gap` applies only between children, so the container's existing `paddingTop` is untouched and the first member doesn't gain a stray offset. Two label-less, note-less members now read as two distinct logged things rather than a paragraph. This is the right shape of fix: the members' own markup is unchanged, and the separation is a property of the group container.

**Accessibility label pluralisation — resolved.** "1 item" / "N items". Small, but a single-member Routine announcing "1 items" is the kind of seam that makes an app feel unattended, and it's now pinned by a test.

**PR description — corrected.** The degradation path now reads "silently falls back", which matches the bare `catch` in `src/hooks/useTraceEntries.ts`. Correcting the prose rather than adding a log is the right call: the swallow is deliberate, the hook comment already explains why, and the user-facing consequence (ungrouped list, history never blanks, no entry hidden) is unchanged and still correct.

**Visible member count on collapsed group rows — correctly deferred.** Agreed this isn't a critic's call to apply. Sarah made an explicit decision to ship without a member-count suffix, and my `· 4 items` note was a suggestion for the ambiguity a filtered group row can carry, not a defect. Flagging it to her separately is the right route. Recording it here so it stays visible: under an active filter, a collapsed group gives no visible signal that only some members matched — muting on expand is the explanation, and it arrives only if the user expands.

### Standing findings from the first pass

Unchanged and still non-blocking:
- Entries logged through a Routine have no path to "Show context", including individually inside an expanded group. Worth revisiting later — each member has one unambiguous timestamp, so the affordance could sit on members without touching the group-row rule.
- Group members render without the entry-type icon, unlike every other Trace row. Worth eyeballing on device alongside the new spacing.

Everything else from the first review stands: all six acceptance criteria on #126 met, empty-state and error copy neutral and non-judgmental, no scores or streaks or completion framing, filter behaviour preserved for the single-Focus case, and no schema change or path by which logged data can be lost.

sfrankle (commented):
## Review feedback — dispositions

All four critics now PASS. Remaining non-blocking findings, and what happened to each.

### Fixed (9a429da)
- Expanded group members had no vertical separation — now `gap: spacing.elementGap` (12dp).
- Accessibility label said "1 items" — now pluralised, test-pinned.
- PR description claimed the enrichment catch logs; it is bare. Description corrected rather than adding a log, since the silent swallow is deliberate and documented in the hook.

### Checked and rejected
- **"The midnight section-ordering edge is documented but untested."** Not accurate — `traceUtils.test.ts:298` covers a completion at 23:50 whose member entry rolled past midnight, asserting the group files under `completedAt`.

  I also probed the ordering hazard the finding gestures at (a group filing back a day landing in a section that then sorts ahead of a newer one) and could not reproduce it. It cannot occur: a group is filed under its `completedAt` date, and any entry that precedes its member in the newest-first stream is necessarily newer than that member, so it files under the same date or a later one. The reordering requires an entry on the later date that is *older* than a 00:0x member — impossible. Probe discarded, no test added, no code change.

### Deferred to Sarah (product decisions, not mine to apply)
- **Visible `· N items` on collapsed group rows.** Under an active filter a collapsed group gives no visible signal that only some members matched; muting-on-expand only explains it if the user expands. Sarah explicitly chose to ship without a member-count suffix, so changing that is hers.
- **Group members render without the entry-type icon** that leads every other Trace row.
- **Routine-logged entries have no "Show context" path**, individually or grouped. Deliberate for groups (no single focal entry for a ±2h window); the individual case is a side effect.

### Deferred as follow-up work (raised in the simplify pass, unchanged here)
- `getRoutineCompletions` re-implements `collapseTraceRows` — a second collapser over the same `EntryTraceRaw` shape, which is the drift the file's own doc comment warns about.
- `matchedIds` puts list-wide filter state inside the render model.
- The unfiltered read re-fetches entry rows it already has; a metadata-only query would do.
- `renderItem` identity churn re-renders every visible row on any expand.

Verification at this commit: `npx tsc --noEmit` clean, eslint 0 errors (10 pre-existing warnings), 824/824 Jest tests passing. The Maestro flow has still not been run — no simulator in this environment.

Issue #125, opened 2026-04-08
Routine time-block visibility and completion tracking


sfrankle:
## Story
As a user, I want Routines to surface on the dashboard when they're due and move out of the way when they're done so that I always know what's relevant right now without being shown things I've already completed.

### Visibility rules
- [ ] A Routine appears as due in the current time block if it is configured for that block and has not yet been completed this block
- [ ] A Routine moves to Completed if either condition is true:
  1. It has been completed at least once today with a timestamp within the current block's window
  2. Total completions today ≥ number of configured time blocks (fully done for the day)
- [ ] A Routine that is not fully completed stays accessible for the rest of the day — no language implying it is overdue or missed

### Examples (Routine configured Morning + Afternoon)
- [ ] Completed at 8:55. At 11:55 (still morning): completed in current block → Completed
- [ ] At 12:05 (afternoon opens): 1 completion < 2 blocks, not yet done in afternoon → Due
- [ ] Completed again at 12:30. At 13:00: 2 completions ≥ 2 blocks → fully done, Completed all day
- [ ] Completed only at 8:55. At 20:00 (past all blocks): 1 < 2, not completed in evening → Due (not overdue)

## Notes
Completion state is derived from `entry` records — no separate completion table. Query: count entries where `routine_id` = this Routine and created_at is today; compare against number of configured time blocks and current block window.

Depends on time block naming investigation (#120) being resolved first.

PR #191, opened 2026-07-29
Derive Routine completion state from the batched day-progress read

### LINKED ISSUE #180: Derive Routine completion state from the batched day-progress read
## Summary

The Tend dashboard currently issues two independent reads of `routine_completion` for the same Routines on the same day. Everything the first needs is a strict subset of the rows the second already fetches.

## Current shape

- `getRoutineCompletionState(db, routineId, timeBlock, today)` — called once per Routine via `Promise.all`, and each call makes up to **3** sequential `getFirstAsync` round trips. So 3N round trips per dashboard focus.
- `getRoutineDayProgress(db, routineIds, today)` — **one** batched query returning `completionCount` and `completedBlocks` for every Routine.

The second already returns everything the first computes. The only extra input the state derivation needs is the configured block count, which is already in memory on `Routine.timeBlocks` — `RoutineSection` uses `routine.timeBlocks.length` for exactly this. `getRoutineCompletionState` re-queries it with `SELECT COUNT(*) FROM routine_time_block`.

## Why it is worth doing

Query count is the lesser reason — N is small and SQLite is local. The real problem is **two independent async sources of truth for one derived fact**. They resolve at different times, so a card's progress line and its group placement can come from two different reads. Deriving one from the other closes that window, and makes state derivation a pure function that is unit-testable without a database.

## Acceptance criteria

- [ ] `RoutineCompletionState` derived as a pure function over `RoutineDayProgress` + `Routine.timeBlocks` + current block
- [ ] One read of `routine_completion` per dashboard refresh, not 3N + 1
- [ ] All four worked examples from #125's AC still pass — these are the regression guard, do not weaken them
- [ ] The `fully_done` boundary (`completions >= blocks`) is preserved; the test pinning the unclamped "3 of 2" invariant added in #178 still passes
- [ ] `src/hooks/useRoutineKeyedRead.ts` consumers collapse to a single read where sensible

## Notes

Deliberately deferred from #178 (PR #178). Doing it there would have reshaped `getRoutineCompletionState` — described in #160 as the most complex piece in the milestone — under a PR that had already cleared three critics and CI. The docstring on `getRoutineDayProgress` records the split as intentional and temporary; update it when this lands.

Related: `docs/decisions.md`, `src/lib/db/queries.ts`, `src/hooks/useRoutineCompletionStates.ts`, `src/hooks/useRoutineDayProgress.ts`.

sfrankle:
## Summary

- The Tend dashboard was asking the database the same question twice. `getRoutineCompletionState` ran up to 3 sequential queries per Routine to work out which group a card belongs in, while `getRoutineDayProgress` already fetched all of it in one batched query for the progress line.
- The real cost wasn't query count — N is small and SQLite is local. It was **two independent async sources of truth for one fact**. They resolved at different moments, so a card's progress line ("1 of 2") and its group placement (due now / completed) could briefly come from different reads and disagree.
- Completion state is now derived as a pure function over the batched read plus `Routine.timeBlocks`, which the component already had in memory. One read of `routine_completion` per refresh, and the two displays are provably consistent because they come from the same resolved rows.

## What to review

- **`src/lib/utils/routine-dashboard.ts`** — the new `deriveRoutineCompletionState` / `deriveRoutineCompletionStates`. Two judgment calls worth a look:
  - The `timeBlocks.length > 0` guard exists so a zero-block "anytime" Routine can't satisfy `0 >= 0` and report itself fully done.
  - `deriveRoutineCompletionStates` **skips** a Routine missing from the progress map rather than defaulting it to `'due'`. That defers to `groupRoutinesForDashboard`, which parks a stateless Routine in "later" — better than presenting an unverified Routine as due now.
- **`src/hooks/useRoutineDayState.ts`** — replaces `useRoutineCompletionStates`, returning progress and states from one read. The `useMemo` is deliberately indifferent to `routines` identity — a pure loop over a handful of items with no I/O, where an id-string dep would cost more indirection than it saves. (`RoutineSection` does memoise `routines`, so the recompute is rarer than the original version of this note claimed.)

- **`src/lib/utils/routine-dashboard.ts`** also now owns `RoutineCompletionState`, moved out of `query-types.ts` during the simplify pass — no query produces or consumes it, and leaving a presentation enum in the query contract preserved the mental model this PR exists to correct.
- **`src/lib/db/query-types.ts`** — the `RoutineDayProgress` docstring previously recorded the split as intentional and temporary (per #178). Updated to say the opposite: new dashboard facts about today's completions belong on this shape, not in a second read.
- **`src/lib/db/queries.ts`** — `getRoutineCompletionState` and its `SELECT COUNT(*) FROM routine_time_block` are gone. No schema change.

## Testing

- The four worked examples from #125's acceptance criteria were kept as the regression guard and not weakened — they now run through `getRoutineDayProgress` + derive instead of the removed query, still database-backed in `queries-routines.test.ts`.
- The `fully_done` boundary (`completions >= blocks`) is preserved. The "three completions against two blocks" test pinning the unclamped "3 of 2" invariant from #178 still passes, and now has a pure twin in `routine-dashboard.test.ts`.
- New pure unit tests for the derivation in `src/lib/utils/__tests__/routine-dashboard.test.ts` — no database needed, which was part of the point.
- `useRoutineCompletionStates.test.ts` replaced by `useRoutineDayState.test.ts`.
- Maestro `routine-dashboard.yaml` gained an assertion on the progress line of the row a completion just moved, so the end-to-end agreement between group placement and progress text is covered. Matched partially (`.*1 of 1.*`) because the trailing block name varies with the wall-clock time the flow runs.

## Checklist
- [x] Acceptance criteria met
- [x] Migration test written if schema changed — N/A, no schema change
- [x] CI passing

Closes #180
Contributes to #125


sfrankle (commented):
## Simplify Pass

Four cleanup reviews (reuse, simplification, efficiency, altitude). Efficiency came back clean — the PR is itself the efficiency fix, and the new `useMemo` deps hold. Applied:

- **`RoutineCompletionState` moved** from `src/lib/db/query-types.ts` to `src/lib/utils/routine-dashboard.ts`. No query produces or consumes it any more; leaving a presentation enum in the query contract preserved the exact mental model this PR exists to correct.
- **Dropped two pure-delegation tests** from `useRoutineDayState.test.ts` (empty-routines short-circuit, query-rejection path). Both restate `useRoutineDayProgress.test.ts` assertions about the same underlying hook, giving one contract two owners. The focus-refresh test stays — it asserts state transitions *through* the derive, which is genuinely this hook's. Test count 836 → 834.
- **`FIXTURE_ROUTINES` now uses a `makeRoutine` factory** instead of two hand-rolled `Routine` literals — the fourth such literal in the suite; `Routine` has already been extended twice.
- **`insertRoutineWithBlocks` returns `{ id, blocks }`** in `queries-routines.test.ts`. The block list was restated at every `completionState` call site, six times over, so editing one copy would silently change what was tested rather than fail.
- **`UseRoutineDayStateResult` now extends `UseRoutineDayProgressResult`** rather than restating `progress`/`loading`/`error`.
- **Corrected a misleading comment** in `useRoutineDayState.ts` claiming `routines` is a new identity every render — `RoutineSection.tsx:57` memoises it. The comment pre-justified churn that is not happening.

### Skipped, deliberately

- **Collapsing the hook layers.** Three reviews independently flagged that `useRoutineKeyedRead` is now a single-caller generic, and the chain is `RoutineSection → useRoutineDayState → useRoutineDayProgress → useRoutineKeyedRead`, one caller each. Real finding, but restructuring the hook module map is beyond this ticket's AC and deserves its own decision.
- **Dropping `useRoutineDayProgress` from the hooks barrel.** Two reviews argue the un-derived read staying public means nothing structurally enforces the single-read invariant. Deferred at plan-review time as outside AC — flagging that the case for it got stronger.
- **Moving the Night→Evening coercion out of `RoutineSection`.** Would change pinned behaviour; the plan holds Night as unchanged, not fixed.
- **Deduplicating the #125 worked examples** across the pure and DB-backed suites. The AC names them the regression guard and says not to weaken them.

`npx tsc --noEmit` clean, `eslint` 0 errors, 834/834 tests pass.

sfrankle (commented):
## Safety Review

**Verdict:** PASS

### Privacy: ✅
No new `fetch`/`axios`/HTTP client, no URLs, no analytics/crash/telemetry imports. No dependency manifest changes (`package.json`/`app.json` are not in the diff), so no new permissions or `expo-*` packages. The only `expo-` reference added is a `jest.mock('expo-router', ...)` in a test.

### Data Safety: ✅
Confirmed: no schema change and no migration in this diff. `src/lib/db/queries.ts` is deletion-only at the SQL level — `getRoutineCompletionState` and its `SELECT COUNT(*) FROM routine_time_block` are removed and nothing replaces them; `getRoutineDayProgress` (the surviving batched read) is untouched apart from its docstring. No `CREATE`/`ALTER`/`DROP`, no writes, no seed inserts. The only `INSERT` statements in the diff are inside test fixtures in `src/__tests__/db/queries-routines.test.ts`. Behaviour that could lose or misreport user data is preserved and pinned: the `completions >= blocks` boundary, the `timeBlocks.length > 0` guard against a zero-block Routine reading as fully done, and the deliberately unclamped "3 of 2" progress invariant.

### Judgmental Language: ✅
Grep for good/bad/great/failed/missed/behind/streak/score/rank/incomplete returns three hits, none user-facing:
- a code comment in `src/hooks/useRoutineDayProgress` usage ("after it failed")
- a test error fixture `new Error('db failed')`
- `docs/decisions.md`, where the words appear in the rule that forbids them in copy

No UI strings, error messages, or empty states changed. The existing copy guard test in `src/lib/utils/__tests__/routine-dashboard.test.ts` survives the refactor.

### Workflow Artifacts: ✅
- `Closes #180` present
- `Contributes to #125` present
- `docs/changelog.md`: exactly one new row, at the top, using PR number #191 (not the issue number), correctly linked

All safety checks passed.

sfrankle (commented):
## Product Vision Review

**Verdict:** PASS

Reviewed against #125 (Routine time-block visibility and completion tracking) and #180. This is a behaviour-neutral refactor, so the bar I held it to was: nothing a user can see or feel has shifted.

### User Story Fulfillment

The four worked examples in #125's AC are the load-bearing part of the story, and they survive intact — still database-backed in `queries-routines.test.ts`, now routed through `getRoutineDayProgress` + derive rather than the removed query. The two rules the story cares about are both preserved in `deriveRoutineCompletionState`:

- completion in the current block wins → `completed_this_block`
- total completions >= configured blocks → `fully_done`

I checked the hour semantics on both sides, since that is where an "equivalent refactor" usually isn't. The removed `timeBlockWindow` in `queries.ts` used `[5,12) [12,14) [14,18) [18,22)`; `getTimeBlock` in `timestamp.ts` uses exactly the same boundaries against the same `substr(created_at, 12, 2)` character positions. Same block, same rows. No user would land in a different group after this change.

The third criterion — a Routine that isn't fully completed stays accessible all day, with no overdue framing — is untouched. `groupRoutinesForDashboard` is unchanged, the disclosure label is unchanged, and nothing new implies a deadline.

### UX Principles

- No new user-facing strings. `disclosureLabel` still returns only `Completed`, `Later`, `Completed · Later`.
- The empty state (no Routines → the `+ Add Routine` row alone) is unchanged and still reads as an invitation, not a gap.
- Error handling is unchanged and still correct for Haven: a failed read degrades to the add row with no alarm surface. The Routine falls into the disclosure rather than vanishing.
- `formatRoutineProgress` still refuses to clamp — "3 of 2" over "2 of 2". Good. Hiding work the user actually did would be the quiet judgment Haven avoids, and the docstring now points at both the pure and the DB-backed test pinning it.

### Interaction

The merged loading flag is the one real timing change, and it lands the right way round. Before, `dueNow` was already gated on the states read resolving, so cards never appeared earlier than they do now — what could happen is a card rendering with a null progress line that filled in a beat later from the second read. That pop-in is now impossible. One read, one moment, both facts. Nothing appears later than it did; one thing stops appearing out of order.

The gate comment on the disclosure block is still accurate post-merge: without state every Routine parks in "later", so flashing an unstated Later row would be actively wrong.

### Side Effects

`useRoutineDayState` has exactly one consumer (`RoutineSection`), and `useRoutineDayProgress` is no longer called anywhere else, so the blast radius is the Tend Routines section and nothing more. No schema change, no migration, no Trace impact. Data-wise nothing is written or deleted — this only changes how existing rows are read.

### Notes (non-blocking)

**Night edge case.** Pinning rather than fixing it is the right call for a refactor PR, and I confirmed it is genuinely preserved: at 23:00 the old code queried the Evening window (18:00–21:59) and missed a 23:00 completion; the new code sees it excluded from `completedBlocks` for the same reason. Identical outcome.

Worth naming the user-facing shape of it, though, because zero-block Routines are creatable today — `useRoutineForm`'s `canSave` requires only a name and valid items, no time block. So a user can make an "anytime" Routine, complete it at 23:30, and watch it read as due again. The completion is not lost (it counts toward the total and shows in the progress line), so nothing disappears — it just doesn't move out of the way, which is precisely the thing #125 exists to guarantee.

It is documented in the `RoutineSection` comment and now in `docs/decisions.md`, but there is no tracking issue for it. Suggest opening one so it doesn't live only in code comments — this is a fix, not a redesign: teaching `getRoutineDayProgress` to record Night, or letting the derivation treat "zero configured blocks + any completion today" as satisfied.

sfrankle (commented):
## Code Quality Review

**Verdict:** PASS

Reviewed `d4c8bfe..5b25ba0`. `npx tsc --noEmit` clean, `eslint` 0 errors (10 pre-existing warnings, none in this diff), `npm test -- --ci` 834/834 across 70 suites.

### Plan-critic concerns — all three honoured

1. **tsc break from a split deletion.** `getRoutineCompletionState`, `timeBlockWindow`, `useRoutineCompletionStates.ts`, the barrel export, and all four consumers (`RoutineSection.tsx`, `RoutineSection.test.tsx`, `TendScreen.test.tsx`, the hook test) all move in one commit, `391f6f0`. No intermediate commit has a dangling import.
2. **`COUNT(*) FROM routine_time_block` ≡ `Routine.timeBlocks.length`.** Covered by the new `the configured block count comes from getRoutines, matching routine_time_block` in `src/__tests__/db/queries-routines.test.ts`, which reads blocks back through `getRoutines(db)`, asserts the length against a raw `SELECT COUNT(*)`, and then feeds those blocks into the derivation. That is exactly the equivalence that was at risk.
3. **`useRoutineDayProgress` barrel export.** Untouched — `src/hooks/index.ts` still exports it. Deferred as asked.

### RN / Expo

- `useMemo` deps in `useRoutineDayState` are complete and honest; the comment explaining why `routines` identity churn is tolerated is the right call for a pure loop over a handful of items.
- Cancellation and focus-refresh stay inside `useRoutineKeyedRead`, unchanged. No new effect without cleanup, no stale closure — `routinesRef`/`loadRef` pattern preserved.
- No new dimensions, colours, or navigation calls.

### TypeScript

- No `any`, no `as` assertions in the new code. `UseRoutineDayStateResult extends UseRoutineDayProgressResult` is a clean widening rather than a restated shape.
- Moving `RoutineCompletionState` out of `query-types.ts` into `routine-dashboard.ts` is correct: no query produces it any more, and the docstring says so.

### Behaviour equivalence (verified, not taken on trust)

- Old: `CAST(substr(created_at, 12, 2) AS INTEGER)` against `timeBlockWindow`. New: `getTimeBlock` → `isoString.slice(11, 13)`. Same 0-based offset, identical boundaries `[5,12) [12,14) [14,18) [18,22)`.
- `getRoutineDayProgress` does **not** restrict `completedBlocks` to a Routine's *configured* blocks, so `completedBlocks.includes(currentBlock)` matches the old unconstrained hour-window query. Had it filtered, this would have been a silent behaviour change.
- The Night edge case is preserved rather than quietly fixed, and is now pinned by an explicit test.

### Tests

- The four #125 worked examples still run against real `routine_completion` rows, now through `getRoutineDayProgress` → derive. Coverage moved, not lost.
- The `12:00:00` boundary test was strengthened: three configured blocks so the `fully_done` check (`1 >= 3`) cannot mask the block comparison. The old two-block version would have passed for the wrong reason.
- Not testing the empty-routines and rejection paths in `useRoutineDayState`, with a comment pointing at `useRoutineDayProgress.test.ts` where they actually live, is the right instinct — I confirmed both cases exist there.
- Maestro assertion is correctly placed inside the expanded disclosure, and `.*1 of 1.*` holds even at Night, when `completedBlocks` is empty and the line degrades to a bare `1 of 1`.

### Haven patterns

- `docs/decisions.md` row appended at the end, with the `>=` boundary, the `timeBlocks.length > 0` guard, and the skip-vs-default-to-due choice all recorded as constraints. The two stale references in the 2026-05-22 and 2026-07-27 rows are corrected rather than left to rot — that is the part most PRs skip.
- One changelog row, PR number linked, `CI passing` correctly left unchecked.

### Non-blocking nits

- `src/lib/utils/__tests__/routine-dashboard.test.ts` — the test named `a Routine with no configured blocks is never fully_done` asserts `completed_this_block`. The assertion is right; the name describes the guard rather than the result. `…is completed_this_block, not fully_done` would read truer.
- `src/components/__tests__/RoutineSection.test.tsx` — `setup()` still takes `statesError` now that there is one error channel. `dayStateError` would match the hook.
- `src/hooks/__tests__/useRoutineDayState.test.ts` — module-level `mockFocusCallback` is not reset in `beforeEach`; `jest.clearAllMocks()` does not clear it. Harmless today (one test uses it, and it is reassigned on every render), but it is a leak waiting for a second test to depend on it.
- `formatRoutineProgress` accepts `RoutineDayProgress | undefined` and handles the gap, while `deriveRoutineCompletionState` requires it and pushes the guard up into `deriveRoutineCompletionStates`. Two conventions for the same missing key on the same map. Defensible as written (the plural wrapper is the only real call site), just worth knowing about.

Code quality looks good.

sfrankle (commented):
## Review Feedback Processed

Three of the four code-quality nits applied in `6480ec6`:

- Test renamed to `a Routine with no configured blocks is completed_this_block, not fully_done` — the old name described the guard, not the assertion.
- `setup()` option `statesError` → `dayStateError`, matching the single error channel the hook now exposes.
- `mockFocusCallback` reset in `beforeEach`. `jest.clearAllMocks()` does not touch module-level state, so a callback left over from a previous test would close over an unmounted hook.

### Pushing back on the fourth

The asymmetry between `formatRoutineProgress` (accepts `RoutineDayProgress | undefined`) and `deriveRoutineCompletionState` (requires it) is intentional, and I would rather not flatten it.

`formatRoutineProgress` returns `string | null` where null means "render nothing", so absorbing a missing key is natural — it already has a way to say "no answer". `deriveRoutineCompletionState` is total by construction: every completion state it can return is a state the card will act on, and there is no null-shaped answer that is not a lie. Making it partial would mean either inventing a return value for "not read yet" or spreading the omit-vs-default decision across two functions. Right now that decision lives in exactly one place — the `if (p === undefined) continue` in `deriveRoutineCompletionStates` — and it is precisely the decision that keeps a missing key from silently becoming `due`. Concentrating it is the point.

Happy to revisit if a second call site for the singular function ever appears.

### Filed from the product-vision review

#192 — Night completions do not move a Routine out of due. I verified the reachability claim before filing: `canSave` in `src/hooks/useRoutineForm.ts:121` has no time-block requirement, so a zero-block "anytime" Routine is user-creatable, and completing it at 23:30 leaves it presenting as due. Pre-existing, deliberately preserved by this PR, now tracked rather than living only in a code comment.

`npx tsc --noEmit` clean, 834/834 tests pass.

